PR #106, opened 2015-05-08
Cache forecast to disk

jakecraige:
This PR implements caches the weather update to disk whenever it's requested. This cache is then used when the app is loaded so the user doesn't have to wait for the update to complete to have something on screen.

I'm opening this as a separate PR than background fetch. This sets up the
prerequisite things we need so I can come back with another PR that uses this
to refresh this data in the background.

In this gif:
1st app start is fresh with no cache.
2nd is with the cache

troposcache

### DIFF
--- a/Tropos/Controllers/TRWeatherController.m
+++ b/Tropos/Controllers/TRWeatherController.m
@@ -8,6 +8,7 @@
 #import "TRDailyForecastViewModel.h"
 #import "TRAnalyticsController.h"
 #import "TRWeatherViewModel.h"
+#import "TRWeatherUpdateCache.h"
 
 @interface TRWeatherController ()
 
@@ -49,9 +50,7 @@
         }];
     }];
 
-    RAC(self, viewModel) = [[[self.updateWeatherCommand.executionSignals switchToLatest] doNext:^(TRWeatherUpdate *update) {
-        [[TRAnalyticsController sharedController] trackEvent:update];
-    }] map:^id(TRWeatherUpdate *update) {
+    RAC(self, viewModel) = [[self latestWeatherUpdates] map:^id(TRWeatherUpdate *update) {
         return [[TRWeatherViewModel alloc] initWithWeatherUpdate:update];
     }];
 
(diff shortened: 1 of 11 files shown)


Euraldius:
This LGTM but it'd be good to get a second opinion @gfontenot @hyperspacemark

Tropos/Controllers/TRWeatherController.m
```diff
@@ -62,15 +59,30 @@ - (instancetype)init
...
+- (RACSignal *)weatherUpdateChanged
+{
+    TRWeatherUpdate *cachedUpdate = [[TRWeatherUpdateCache new] latestWeatherUpdate];
+    RACSignal *weatherUpdates = [self.updateWeatherCommand.executionSignals startWith:[RACSignal return:cachedUpdate]];
+
+    return [[[[weatherUpdates switchToLatest] filter:^BOOL(TRWeatherUpdate *update) {
+        return update != nil;
+    }] doNext:^(TRWeatherUpdate *update) {
+        [[TRAnalyticsController sharedController] trackEvent:update];
+        [[TRWeatherUpdateCache new] archiveWeatherUpdate:update];
```
gfontenot: should these be done via other subscriptions to this signal?

Tropos/Models/TRWeatherUpdate.m
```diff
@@ -68,6 +72,29 @@ - (void)updateCurrentTemperaturesWithConditions:(NSDictionary *)conditions withF
...
+- (instancetype)initWithCoder:(NSCoder *)coder
+{
+    CLPlacemark *placemark = [coder decodeObjectForKey:TRPlacemarkKey];
+    NSDictionary *currentConditions = [coder decodeObjectForKey:TRCurrentConditionsKey];
+    NSDictionary *yesterdaysConditions = [coder decodeObjectForKey:TRYesterdaysConditionsKey];
+
+    return [self initWithPlacemark:placemark currentConditionsJSON:currentConditions yesterdaysConditionsJSON:yesterdaysConditions];
+}
+
+
```
gfontenot: remove extra newline

gfontenot:
one question about side effects, one tiny style nitpick. But this looks pretty good to me.

Tropos/Controllers/TRWeatherController.m
jakecraige: I think that makes sense. How about https://github.com/thoughtbot/Tropos/commit/8f82761ebc8fc0f3a95ec9af43bcbbbf0c1b38a7 ?

gfontenot:
Yeah, that makes more sense to me. Maybe wait for feedback from @hyperspacemark?

jakecraige:
@gfontenot Discussed with @cpytel and we decided to merge so we weren't blocked and could continue making progress. If there are any other concerns we can address them in a separate PR.

PR #123, opened 2015-06-19
Switch from Travis to Circle

jakecraige:
Circle has a newer version of Xcode and generally works better.
Overall I prefer it to Travis.


jakecraige:
Travis will never pass because it decided it wants to endlessly freeze from now on.

Euraldius:
I like Circle :shipit:

PR #124, opened 2015-06-20
Background Fetch

jakecraige:
When the user allows us to always retrieve their location,
we will perform background fetches when the system allows.
If they deny this and set it to "when in use",
the app will disable background fetch and continue to work.

- Switch location services from when in use to always
- Enabled background fetch
- Fetch and cache weather update in background

https://trello.com/c/dNA4MJGZ/5-background-fetching

### DIFF
--- a/Resources/Other-Sources/en.lproj/InfoPlist.strings
+++ b/Resources/Other-Sources/en.lproj/InfoPlist.strings
@@ -1,2 +1,3 @@
 "CFBundleDisplayName" = "Tropos";
 "NSLocationWhenInUseUsageDescription" = "Tropos will only use your location to discover current weather conditions and forecasts for your area.";
+"NSLocationAlwaysUsageDescription" = "Tropos will only use your location to stay updated on current weather conditions and forecasts for your area.";
--- a/UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
+++ b/UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
@@ -1,0 +1,63 @@
+@import CoreLocation;
+#import "TRApplicationController.h"
+#import "TRWeatherController.h"
+#import "TRLocationController.h"
+
+@interface TRApplicationController (Tests)
+
+@property (nonatomic) TRWeatherController *weatherController;
+@property (nonatomic) TRLocationController *locationController;
+
+@end
+
+@interface TRWeatherController (Tests)
+
+@property (nonatomic) RACCommand *updateWeatherCommand;
+
+@end
+
+SpecBegin(TRApplicationController)
+
+TRLocationController* (^locationControllerWithAuthorizationStatusAuthorizedAlwaysEqualTo) (BOOL) = ^TRLocationController* (BOOL enabled){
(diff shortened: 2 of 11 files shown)


jakecraige:
I'll probably look into adding tests soon but I'd still love some feedback on the implementation.

Resources/Other-Sources/en.lproj/InfoPlist.strings
```diff
@@ -1,2 +1,3 @@
 "CFBundleDisplayName" = "Tropos";
 "NSLocationWhenInUseUsageDescription" = "Tropos will only use your location to discover current weather conditions and forecasts for your area.";
+"NSLocationAlwaysUsageDescription" = "Tropos will only use your location stay updated on current weather conditions and forecasts for your area.";
```
cdzombak: > …your location stay updated…

You're missing a "to" in there.

jakecraige: Good catch. Added in https://github.com/thoughtbot/Tropos/commit/ae34fc5121ffd07560e2d2389a4e03ad5edfc403

UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
```diff
@@ -0,0 +1,61 @@
...
+            [ctrl setMinimumBackgroundFetchIntervalForApplication:app];
+
+            OCMVerify([app setMinimumBackgroundFetchInterval:UIApplicationBackgroundFetchIntervalMinimum]);
+        });
+
+        it(@"sets the interval to never when authorization is not always", ^{
+            UIApplication *app = OCMClassMock([UIApplication class]);
+            TRLocationController *locationController = OCMPartialMock([[TRLocationController alloc] init]);
+            OCMStub([locationController authorizationStatusEqualTo:kCLAuthorizationStatusAuthorizedAlways]).andReturn(NO);
+            TRApplicationController *applicationController = [TRApplicationController new];
```
cpytel: In the the spec above you call this same thing `ctrl`. What do you think about making the variables in the setups, since they are identical, named the same thing?

cpytel:
One comment on some variable naming, otherwise good to merge at your discretion. I had fun pairing on this with you while I was in SF, thanks!

UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
jakecraige: Can you clarify what you mean by "the steps"? Do you mean to define a `beforeEach`?

cpytel: Sorry, I meant in the setup for `"sets the interval to minimum when authorization is always"` this variable is `ctrl` but in the setup for `"sets the interval to never when authorization is not always"` the variable is `applicationController`.

and I was suggesting that they could be named the same name.

jakecraige: Ah! Yeah I can do that.

PR #126, opened 2015-06-22
Move to CocoaPods-Keys

### LINKED ISSUE #125: Consider using CocoaPods-Keys
We faced similar problems keeping keys secret when open sourcing our app [Eidolon](https://github.com/artsy/eidolon), so @orta made [CocoaPods-Keys](https://github.com/orta/cocoapods-keys), a CocoaPods plugin designed specifically for the use case you have here.

If you'd like to see what using CocoaPods-Keys instead of Secrets.h would look like, I'd be happy to send a pull request.

ashfurrow:
Previously, API keys were stored in `Secrets.h`, a file not checked into git. This pull requests moves to [CocoaPods-Keys](https://github.com/orta/cocoapods-keys), a CocoaPods plugin designs specifically for this use case. 

This pull requests:

- Updates CocoaPods for use with the new plugin architecture.
- Installs CocoaPods-Keys gem.
- Updates `setup` script to stub Mixpanel and Hockey keys.
- Switches to new keys.
- Migrates CI instructions to use new keys plugin.
- Updates installation instructions.

I've kept `Secrets.h` in `.gitignore` to prevent a developer accidentally checking in this file and leaking secrets. 

Fixes #125.


Podfile.lock
```diff
@@ -20,6 +21,7 @@ PODS:
 DEPENDENCIES:
   - Expecta
   - HockeySDK (~> 3.6.2)
+  - Keys (from `/Users/ash/bin/Tropos/Pods/CocoaPodsKeys/`)
```
jakecraige: Since this references your machine, does this work when other people clone down the app?

ashfurrow: Hey, good catch! This bug was actually just [reported today](https://github.com/orta/cocoapods-keys/issues/92). I'll update the PR with the fixed version.

ashfurrow: Done'd!

jakecraige:
I'm not sure totally sure if we should go this direction for a few reasons:

1. It adds what seems like a lot of extra complexity to hide a few keys that aren't really that secret. If they ended up leaked it wouldn't be that big of deal. The current setup we have using the `Secrets.h` is simple and pretty easy to understand.
2. We're planning on slowly introducing more Swift into the codebase and with that likely is migrating to Carthage. Using CocoaPods-Keys adds a hard dependency on keeping CocoaPods around.

That being said I'd love to hear others thoughts on this. I know @gfontenot has some experience with this so he could provide some real-world experiences to the conversation.

gfontenot:
FWIW, CocoaPods handles Swift just fine, so our decision around dependency management isn't really affected by the language we're choosing.

That being said, I do very much dislike the idea of coupling private key handling with dependency management, so I'm in agreement with @jakecraige about not introducing this change.

ashfurrow:
As you like. Key storage isn't a critical part of our toolchain; if CocoaPods disappeared tomorrow, Artsy would be fine and we would probably do something like this missing file approach. 

If you feel this is a better direction for your team, sounds good 👍 You know what's best for your project.

jakecraige:
Based on this discussion I'm going to close this for now. If we find later that this would be beneficial we can re-evaluate it then.

PR #127, opened 2015-06-22
[TRAppDelegate] Enable auto crash reports to avoid missing out on cra…

bogren:
…sh reports


bogren:
I could not send report due to the app crashed again before the prompt, also I personally prefer to get all reports :) 

In addition I'm working on a bug fix to avoid the crash I experience.

jakecraige:
Thanks! Looks good. I'll merge this sometime this week.

Issue #128, opened 2015-06-23
CocoaPods Error On Build


epmatsw:
I'm trying to build the app on my laptop, but I'm running into an issue with bin/setup.

It looks like the gemfile installs 0.35.0, but the pods file actually requires 0.37+. Running `bundle update` followed by `./bin/setup` appears to fix the error. This is using ruby 2.2.1p85 on OSX 10.11.

```
Fetching gem metadata from https://rubygems.org/.........
Fetching version metadata from https://rubygems.org/...
Fetching dependency metadata from https://rubygems.org/..
Installing i18n 0.6.11
Using json 1.8.1
Installing minitest 5.4.3
Installing thread_safe 0.3.4
Installing tzinfo 1.2.2
Installing activesupport 4.1.8
Installing claide 0.7.0
Installing fuzzy_match 2.0.4
Installing nap 0.8.0
Installing cocoapods-core 0.35.0
Installing cocoapods-downloader 0.8.0
Installing cocoapods-plugins 0.3.2
Installing netrc 0.7.8
Installing cocoapods-trunk 0.4.1
Installing cocoapods-try 0.4.2
Using colored 1.2
Installing escape 0.0.4
Installing molinillo 0.1.2
Installing open4 1.3.4
Installing xcodeproj 0.20.2
Installing cocoapods 0.35.0
Installing xcpretty 0.1.7
Using bundler 1.10.4
Bundle complete! 2 Gemfile dependencies, 23 gems now installed.
Use `bundle show [gemname]` to see where a bundled gem is installed.
Analyzing dependencies
[!] The version of CocoaPods used to generate the lockfile (0.37.2) is higher than the version of the current executable (0.35.0). Incompatibility issues may arise.
/Users/will/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/pathname.rb:430:in `open': No such file or directory @ dir_initialize - /Users/will/.cocoapods/repos (Errno::ENOENT)
	from /Users/will/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/pathname.rb:430:in `foreach'
	from /Users/will/.rvm/rubies/ruby-2.2.1/lib/ruby/2.2.0/pathname.rb:430:in `children'
	from /Users/will/.rvm/gems/ruby-2.2.1/gems/cocoapods-0.35.0/lib/cocoapods/sources_manager.rb:63:in `all'
[...]

PR #129, opened 2015-06-23
Update to use latest versions of gems

epmatsw:
Specifically, the cocoapods gem that the Gemfile.lock specified was 0.35, while the build process actually required 0.37. This should fix that as well as pulling in updated versions of the other gems.

Note that the tests are failing in the branch (at least on my machine). However, I can't tell if that's a pre-existing issue or if it's caused by this since I can't get the project to build without this change. If someone with a working build environment could chime in one way or the other, that would be nice.


epmatsw:
Fixes #128

jakecraige:
Looks great. CI is passing which is nice. I'll pull it down to make sure it's still passing locally and then merge. Thanks!

jakecraige:
Looks good! When I pulled down locally and after I `bundle install`d, I got some linker errors but once I `pod install`d it all worked. Maybe that'll fix your local errors?

jakecraige:
Merged in https://github.com/thoughtbot/Tropos/commit/4dd7a6eddc0a643f2ca33145738325192e2e7694

PR #131, opened 2015-06-26
Cache the weather update date

jakecraige:
We want to cache the date along with the weather update
since this is used to display "Last updated at".

Without this it will always say it was last updated at the current time
and when the real update comes in it won't change
but the new weather will.


cpytel:
This looks good to merge.

PR #133, opened 2015-06-30
Add TRTemperatureComparisonFormatter+initWithDate for context aware comparisons

jakecraige:
There was a bug where comparisons
shown from the cache would show inaccurate text
because it wasn't using the date from the cache.

Ex:
If the date was cached in the AM and you open the app in the afternoon it might say,
"Last updated 8:00AM, It's the same this afternoon as yesterday afternoon."
when it should say,
"Last updated 8:00AM, It's the same this morning as yesterday morning."

This PR has the comparison be initialized with a date
that it uses when checking the time of day.

I chose to switch from a class method to instance method
so that I didn't have to pass the date
through a chain of class methods.

### DIFF
--- a/Tropos.xcodeproj/project.pbxproj
+++ b/Tropos.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		130C4C29D1C62D52D0767A6B /* Main.storyboard in Sources */ = {isa = PBXBuildFile; fileRef = E59B090E67C8BD269D3CD9DB /* Main.storyboard */; };
+		23AB769C1B4222830083EC14 /* TRTemperatureComparisonFormatterSpec.m in Sources */ = {isa = PBXBuildFile; fileRef = 23AB769B1B4222830083EC14 /* TRTemperatureComparisonFormatterSpec.m */; };
 		46660A27B04F4C7B78A7920F /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 436D9431497525313B32E225 /* Foundation.framework */; };
 		4D0A5C381A3A27510084C41E /* TRForecastController.m in Sources */ = {isa = PBXBuildFile; fileRef = 4D0A5C371A3A27510084C41E /* TRForecastController.m */; };
 		4D0A5C3D1A3A53DD0084C41E /* CWForecastDetailView.xib in Resources */ = {isa = PBXBuildFile; fileRef = 4D0A5C3C1A3A53DD0084C41E /* CWForecastDetailView.xib */; };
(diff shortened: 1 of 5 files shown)


Tropos.xcodeproj/project.pbxproj
```diff
@@ -198,7 +200,6 @@
 		DED70296CC99B4AAD0DE42BD /* Pods.release.xcconfig */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = text.xcconfig; name = Pods.release.xcconfig; path = "Pods/Target Support Files/Pods/Pods.release.xcconfig"; sourceTree = "<group>"; };
 		E2B218FB5F097AAE3080F4DD /* UnitTests-Prefix.pch */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = sourcecode.c.h; path = "UnitTests-Prefix.pch"; sourceTree = "<group>"; };
 		E59B090E67C8BD269D3CD9DB /* Main.storyboard */ = {isa = PBXFileReference; includeInIndex = 1; lastKnownFileType = file.storyboard; path = Main.storyboard; sourceTree = "<group>"; };
-		E799194B1ACEE75C00A5600C /* Secrets.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = Secrets.h; sourceTree = "<group>"; };
```
tonyd256: why is this removed here?

jakecraige: It probably shouldn't. It keeps happening on my comp and I'm not sure why. I'll undo it.

tonyd256:
@jakecraige what do you mean by "through a chain of class methods." Would it be fine to just pass the date into the class method?

Tropos.xcodeproj/project.pbxproj
tonyd256: probably because it's in the .gitignore or something

jakecraige:
@tonyd256 Because `localizedStringFromComparison:adjective:precipitation` ends up calling other class methods I would have to pass the date through the entire chain of them. 

Mainly it's because `timeOfDay` needs the date and it's being called form a few places, so I'd also have to pass in the date to those methods(`localizedCurrentTimeOfDay` and `localizedPreviousTimeOfDay`). 

They would have to accept the date and pass it through to `timeOfDay`. I could also perhaps pass in a `TRTimeOfDay` so it doesn't have to call `timeOfDay`. 

I could give it a shot like that if you think it'd be preferable.

tonyd256:
Makes sense :+1:

jakecraige:
Closing for #144

PR #139, opened 2015-07-02
Classes from Models directory converted to Swift

tomkowz:
Hey,

Finished working on converting model to Swift. It took me a while to convert `TRWeatherUpdate` which has pretty complicated initialization. I don't like how JSON parsing stuff is done in Swift because of necessary of casting a lot everywhere. I split up yesterdays conditions and current conditions (will be easy extendable in the future) and passing new `WeatherConditions` object to `WeatherUpdate` during initialization.

I also created enums for NSCoding stuff to avoid word mistakes when working on NSCoding stuff. I don't like boilerplate that is needed for implementing this, but, didn't find better solution.

If there are any thing that might be better coded just let me know and we'll discuss it ;)

Thanks!


jakecraige:
Wow. Thanks for doing all this work. It's great to see someone so interested in helping out and we really appreciate it.

We generally like to keep PRs fairly small so that we can thoroughly review them and really understand what's going on. Large PRs tend to be more difficult to review and stay open longer because of the amount of changes they make.

Would you be open to splitting this up into separate PRs? The commits you made provide a nice separation you could use.

tomkowz:
Sure, I can, will do It When back home :) thanks for suggestion. I thought it might be too big change.

Br, Tomasz Szulc
+48 788 434 094
http://szulctomasz.com


Dnia 2 lip 2015 o godz. 18:42 Jake Craige <notifications@github.com> napisał(a):

PR #140, opened 2015-07-02
Convert TRTemperature and TRDailyForecast to Swift

tomkowz:
Just converted these classes to Swift, added tests.


jakecraige:
I don't mean to be a pain, but could you split this into 2 separate PRs? Start with only 001ed982356b4a540d52e404d0731a310e2eab57 and we can look at it and move forward from there

PR #141, opened 2015-07-02
Convert TRTemperature to Temperature

tomkowz:
Converted class to Swift.

PR #143, opened 2015-07-02
Convert TRDailyForecast to DailyForecast

tomkowz:
Please merge it after #141 is finally merged.


tomkowz:
I'll create another pull request for this because I created it too early and there was changes in #141.

PR #144, opened 2015-07-03
TRTemperatureComparison+localizedStringFromComparison accepts a date

jakecraige:
There was a bug in where comparisons
shown from the cache would show inaccurate text
because it wasn't using the date from the cache.

Ex:
If the date was cached in the AM and you open the app in the afternoon it might say,
"Last updated 8:00AM, It's the same this afternoon as yesterday afternoon."
when it should say,
"Last updated 8:00AM, It's the same this morning as yesterday morning."

This PR has the comparison be initialized with a date
that it uses when checking the time of day.

I chose to switch from a class method to instance method
so that I didn't have to pass the date
through a chain of class methods.

https://trello.com/c/IeLJK4C8/36-temperature-comparison-text-not-taking-cached-date-into-account

### DIFF
--- a/UnitTests/Tests/Formatters/TRTemperatureComparisonFormatterSpec.m
+++ b/UnitTests/Tests/Formatters/TRTemperatureComparisonFormatterSpec.m
@@ -1,0 +1,34 @@
+#import "UnitTests-Swift.h"
+#import "TRTemperatureComparisonFormatter.h"
+
+@interface TRTemperatureComparisonFormatter (Tests)
+
++ (NSCalendar *)calendar;
+
+@end
+
+SpecBegin(TRTemperatureComparisonFormatter)
+
+NSDate* (^dateFromString) (NSString*) = ^NSDate *(NSString *dateString) {
+    NSDateFormatter *dateFormatter = [NSDateFormatter new];
+    [dateFormatter setDateFormat:@"yyyy-MM-dd HH:mm:ss zzz"];
+    return [dateFormatter dateFromString:dateString];
+};
+
+describe(@"TRTemperatureComparisonFormatter", ^{
+    describe(@"localizedStringFromComparison:adjective:precipitation", ^{
+        it(@"bases the time of day off of the date", ^{
+            [[TRTemperatureComparisonFormatter calendar] setTimeZone:[NSTimeZone timeZoneWithAbbreviation:@"UTC"]];
+            NSString *adjective;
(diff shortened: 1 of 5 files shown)


jakecraige:
@tonyd256 What do you think about this implementation instead of #133. It's the one where I pass the date around.

tonyd256:
honestly this isn't that bad. I think it makes more sense, especially readying the function signatures. My vote would be for this instead of #133

jakecraige:
@tonyd256 @gfontenot Can I get one another once-over / review on this? Cleaned up the merge conflicts. Tests are good locally.

UnitTests/Tests/Formatters/TRTemperatureComparisonFormatterSpec.m
```diff
@@ -0,0 +1,34 @@
...
+
+@interface TRTemperatureComparisonFormatter (Tests)
+
++ (NSCalendar *)calendar;
+
+@end
+
+SpecBegin(TRTemperatureComparisonFormatter)
+
+NSDate* (^dateFromString) (NSString*) = ^NSDate* (NSString *dateString) {
```
gfontenot: Can you move these pointers to the other side of the space? (`NSDate *(` instead of `NSDate* (`)

gfontenot: Actually, we should have Quick/Nimble/Swift in master at this point, right? Maybe we can test this whole thing in Swift and not have to deal with this at all?

gfontenot:
one comment re: testing, but this looks good to me.

UnitTests/Tests/Formatters/TRTemperatureComparisonFormatterSpec.m
jakecraige: No problem on the pointers.

Because of the way I'm testing this right now I don't think we can. To keep a consistent time zone in test I've made a category up on LN4 to expose the private calendar, then on LN21 I set it's time zone to a specific one. There's likely other ways to accomplish this but for now this does the trick. We can probably set up the class a little better for testing when converting it to Swift.

jakecraige:
> Test Suite 'All tests' started at 2015-08-24 16:02:07 +0000
> Test Suite 'All tests' passed at 2015-08-24 16:02:07 +0000.
> 	 Executed 28 tests, with 0 failures (0 unexpected) in 0.054 (0.087) seconds

PR #145, opened 2015-07-03
Convert TRPrecipitation to Precipitation.

tomkowz:
Converted class and removed unused TRWeatherLocation.


UnitTests/Tests/PrecipitationSpec.swift
```diff
@@ -0,0 +1,26 @@
...
+import Quick
+import Nimble
+
+class PrecipitationSpec: QuickSpec {
+    override func spec() {
+        describe("Precipitation") {
+            context("precipitation chance") {
+                it("returns none for 0% chance of precipitation") {
+                    let precipitation = Precipitation(probability: 0.0, type: "")
+                    expect{precipitation.chance}.to(equal(PrecipitationChance.None))
```
jakecraige: Should these expect's use `expect()` instead of `expect{}` ?

tomkowz: Oh, didn't see that. There is trailing closure syntax in this function so this is fine by using this but this was accident that I did it. We can leave it as is or change if you want.

Tropos/Models/Precipitation.swift
```diff
@@ -0,0 +1,23 @@
...
+    case None, Slight, Good
+}
+
+@objc class Precipitation: NSObject {
+    let type: String
+    let probability: Float
+    
+    var chance: PrecipitationChance {
+        switch probability {
+        case probability where probability > 30.0: return .Good
```
jakecraige: Thoughts on these being `case _ where` since we aren't using the probability here?

tomkowz: Yeah, nice catch.

jakecraige:
@tomkowz Rather than overwriting the changes in a PR. It's helpful to us if you make separate commits for the changes because it's easier to see what changes you made. 

When we merge it we'll squash it back down to a single or few commits anyways so you don't have to worry about it.

tomkowz:
Okay, thanks for suggestion, so you're responsible for clearing things up after PR is accepted? I thought I should clean it up but wasn't sure.

tomkowz:
The same crash reason as earlier.

jakecraige:
> so you're responsible for clearing things up after PR is accepted?

Well... to a point of course, lol. :) 

Ideally when it's first opened it has a clean history with good commit messages and then as you make changes in response to reviews it's nice to see them as separate commits.

 I have an open that I think is a good example [here](https://github.com/thoughtbot/guides/pull/342/commits). There's an initial commit with a descriptive commit message and then each subsequent commit is small changes based on the reviews. When merging I do a fixup on the last 3 commits and it's done.

We merge on the command-line anyways so it's not a big deal to squash/fixup extra commits as needed.

tomkowz:
@jakecraige, okay, understood, sound good to me. Will remember next time ;)

jakecraige:
Yeah I'm not sure what's going on with Circle... It was working fine briefly and now stopped. 

Master isn't even passing. It's passing on Ship.io now. Unfortunately I don't think Ship.io runs PRs.
<a href='https://app.ship.io/dashboard#/jobs/8374/history' target='_blank'></a>

tomkowz:
@jakecraige I will take a look what's wrong with these `UIColor_TRTemperatureColorSpec` tests later the day and maybe found something.

jakecraige:
To me clear, I don't think it's your code. Do you see something that makes you think that test is behaving funny? 

I sent a support ticket into Circle to see if they can provide any insight since Ship.io seems to be working fine with the same setup.

tomkowz:
@jakecraige sure, I now this is not my code ;)

I looked in this test and it is very simple but crash goes from PhotoLibrary private framework and this is very weird because app doesn't use it.

jakecraige:
Yeah I noticed the same thing. It's super weird. 

Either way this looks good. Merged in https://github.com/thoughtbot/Tropos/commit/9672414d4ff0d237a7750862e0606a9e1903a578

tomkowz:
Thanks!

PR #148, opened 2015-07-03
Convert TRDailyForecast to DailyForecast

tomkowz:


tomkowz:
Again something crashed, not crashing locally :/

Tropos/Models/DailyForecast.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+@objc class DailyForecast: NSObject {
+    let date: NSDate
+    let conditionsDescription: String
+    let highTemperature: Temperature
+    let lowTemperature: Temperature
+    
+    init(json: NSDictionary) {
```
jakecraige: How about capitalizing `json` so it's `JSON`?

tomkowz: I can capitalize it but have no opinion. I usually try to keep parameters lowercased but here is json which is always capitalized as shortcut, so, okay, let's do it.

tomkowz: @jakecraige I will also change JSON type to `Dictionary<String, AnyObject>` to keep it more Swifty.

tomkowz:
@jakecraige one question about github, are you notified by github when I push new commit to this my pull request? Asking because don't know if I should mention you when I push some changes according to review.

Tropos/Models/DailyForecast.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+@objc class DailyForecast: NSObject {
+    let date: NSDate
+    let conditionsDescription: String
+    let highTemperature: Temperature
+    let lowTemperature: Temperature
+    
+    init(JSON: Dictionary<String, AnyObject>) {
```
jakecraige: What do you think about turning this into a [failable initializer](https://developer.apple.com/library/mac/documentation/Swift/Conceptual/Swift_Programming_Language/Initialization.html#//apple_ref/doc/uid/TP40014097-CH18-ID224) and doing an if-let `as?` for the JSON values so we can remove the explicit unwraps?

jakecraige: That might also allow you to get rid of the extra stuff you had to add to the tests to get it passing.

jakecraige:
@tomkowz Typically we try to comment on each individual review and I will get notified of that. I don't for pushes though. If you are replying to a comment about it, no need, if you aren't then you can ping me in a new comment.

Also, I remembered we have a whole guide on code review that explains some things the things I've re-typed before [here at thoughtbot/guides](https://github.com/thoughtbot/guides/tree/master/code-review).

Tropos/Models/DailyForecast.swift
tomkowz: If we decide to change it into failable initializer it will require changes in the app logic. `TRWeatherUpdate` always enumerate from 1 to 4 and creates 3 `DailyForecast`s. If we start returning nil there instead of object additional safety checks will be required. I am okay for now to change this explicit unwrap into if-let but I would like to keep initializer not failable because I don't know the app logic on controllers level where RAC is. 

Let's discuss it. 

1. What if something goes wrong and view `DailyForecast` cannot be created. Do we want to display then 2 or just 1 `DailyForecast`? 
2. Can we assume that `DailyForecast` is always created and there is no case when something goes wrong? Then we can use only if-let and initializer will be always returning some objects.

tomkowz:
thanks for this guide!

Tropos/Models/DailyForecast.swift
tomkowz: I do not like failable initializers with classes. The code gets messy so far. You have to fill all stored properties before returning nil. Example for code "improvement".

    @objc class DailyForecast: NSObject {
        let date: NSDate
        let conditionsDescription: String
        let highTemperature: Temperature
        let lowTemperature: Temperature
        
        init?(JSON: Dictionary<String, AnyObject>) {
            if let time = JSON["time"] as? NSTimeInterval {
                self.date = NSDate(timeIntervalSince1970: time)
            } else {
                self.date = NSDate()
            }
            
            if let conditionsDescription = JSON["icon"] as? String {
                self.conditionsDescription = conditionsDescription
            } else {
                self.conditionsDescription = ""
            }
            
            if let temperatureMax = JSON["temperatureMax"] as? Int {
                self.highTemperature = Temperature(fahrenheitValue: temperatureMax)
            } else {
                self.highTemperature = Temperature(fahrenheitValue: 0)
            }
            
            if let temperatureMin = JSON["temperatureMin"] as? Int {
                self.lowTemperature = Temperature(fahrenheitValue: temperatureMin)
            } else {
                self.lowTemperature = Temperature(fahrenheitValue: 0)
            }
        
            super.init()
            if self.dynamicType.validate(JSON) == false { return nil }
        }
        
        class func validate(JSON: Dictionary<String, AnyObject>) -> Bool {
            // json validation goes here
            // check if all necessary values are inside
            return false
        }
    }

What do you think about it @jakecraige? Can we keep it as is for now and think about it more later when we will be able e.g. to convert `DailyForecast` to be struct instead of class? Then we can implement failable intializer even simpler without filling stored properties. It works different for structs.

jakecraige: Great points.

> What if something goes wrong and view DailyForecast cannot be created. Do we want to display then 2 or just 1 DailyForecast?

That's what I'm thinking. After some playing around to get one to not show up I have it looking like this which seems reasonable.

ios simulator screen shot jul 3 2015 8 14 58 am

> Can we assume that DailyForecast is always created and there is no case when something goes wrong? Then we can use only if-let and initializer will be always returning some objects.

I don't think we can. I haven't seen a case where it doesn't work yet but it's always a possibility since we're relying on a third party API for the data. With that in mind we should definitely have some sort of fallback. I'm thinking the easiest way to handle this would be to use the null-object pattern. When it fails we can provide a null-object and it will render empty like in the screenshot I posted earlier. 

I was originally considering making the values optional but that's not really great because there's no case where this object is valid if it's missing any of them. We can't display it. With that in mind I'm thinking there should be a `NullDailyForecast` class that returns `nil` values for everything. This way we don't have to add lots of conditional logic or mess with any RAC stuff.

After some playing around I got a diff that looks like this, still needs the failable initializer stuff:

```diff
diff --git a/Tropos/Models/DailyForecast.swift b/Tropos/Models/DailyForecast.swift
index 0075d48..99c5f25 100644
--- a/Tropos/Models/DailyForecast.swift
+++ b/Tropos/Models/DailyForecast.swift
@@ -13,3 +13,10 @@ import Foundation
         self.lowTemperature = Temperature(fahrenheitValue: JSON["temperatureMin"] as! Int)
     }
 }
+
+@objc class NullDailyForecast: NSObject {
+    let date: NSDate? = .None
+    let conditionsDescription: String? = .None
+    let highTemperature: Temperature? = .None
+    let lowTemperature: Temperature? = .None
+}
diff --git a/Tropos/Models/TRWeatherUpdate.m b/Tropos/Models/TRWeatherUpdate.m
index 56ce074..dadecd5 100644
--- a/Tropos/Models/TRWeatherUpdate.m
+++ b/Tropos/Models/TRWeatherUpdate.m
@@ -54,6 +54,10 @@
 
     for (NSUInteger index = 1; index < 4; index++) {
         DailyForecast *dailyForecast = [[DailyForecast alloc] initWithJSON:currentConditionsJSON[@"daily"][@"data"][index]];
+        if (index == 3) { dailyForecast = nil; } // ONLY FOR TESTING
+        if (dailyForecast == nil) {
+            dailyForecast = (DailyForecast *)[NullDailyForecast new];
+        }
         [dailyForecasts addObject:dailyForecast];
     }
 
diff --git a/Tropos/ViewModels/TRDailyForecastViewModel.m b/Tropos/ViewModels/TRDailyForecastViewModel.m
index bb068a7..2bbd4db 100644
--- a/Tropos/ViewModels/TRDailyForecastViewModel.m
+++ b/Tropos/ViewModels/TRDailyForecastViewModel.m
@@ -30,19 +30,31 @@
     return [dateFormatter stringFromDate:self.dailyForecast.date];
 }
 
-- (UIImage *)conditionsImage
+- (nullable UIImage *)conditionsImage
 {
-    return [UIImage imageNamed:self.dailyForecast.conditionsDescription];
+    NSString *conditionsDescription = [NSString stringWithFormat:@"%@", self.dailyForecast.conditionsDescription];
+    return [UIImage imageNamed:conditionsDescription];
 }
 
-- (NSString *)highTemperature
+- (nullable NSString *)highTemperature
 {
-    return [[TRTemperatureFormatter new] stringFromTemperature:self.dailyForecast.highTemperature];
+    return [self stringFromTemperature:self.dailyForecast.highTemperature];
 }
 
-- (NSString *)lowTemperature
+- (nullable NSString *)lowTemperature
 {
-    return [[TRTemperatureFormatter new] stringFromTemperature:self.dailyForecast.lowTemperature];
+    return [self stringFromTemperature:self.dailyForecast.lowTemperature];
+}
+
+#pragma mark - Private
+
+- (NSString *)stringFromTemperature:(Temperature *)temperature
+{
+    if (temperature == nil) {
+        return @"";
+    } else {
+        return [[TRTemperatureFormatter new] stringFromTemperature:self.dailyForecast.lowTemperature];
+    }
 }
 
 @end
```

Thoughts?

jakecraige: I just saw your comment. Reading it now

tomkowz: Ok, let me know when you're done and I will re-read it ;)

jakecraige: Interesting. I didn't realize that you have to initialize all values on a class. That's pretty frustrating and I see how it would be a pain. Learned something new today.

What about something like this? Pretty much defining our own style of failable initializer

```diff
diff --git a/Tropos/Models/DailyForecast.swift b/Tropos/Models/DailyForecast.swift
index 0075d48..b9c9acd 100644
--- a/Tropos/Models/DailyForecast.swift
+++ b/Tropos/Models/DailyForecast.swift
@@ -1,15 +1,33 @@
 import Foundation
 
-@objc class DailyForecast: NSObject {
+@objc protocol Forecastable { }
+
+@objc class DailyForecast: Forecastable {
     let date: NSDate
     let conditionsDescription: String
     let highTemperature: Temperature
     let lowTemperature: Temperature
     
-    init(JSON: Dictionary<String, AnyObject>) {
-        self.date = NSDate(timeIntervalSince1970: JSON["time"] as! NSTimeInterval)
-        self.conditionsDescription = JSON["icon"] as! String
-        self.highTemperature = Temperature(fahrenheitValue: JSON["temperatureMax"] as! Int)
-        self.lowTemperature = Temperature(fahrenheitValue: JSON["temperatureMin"] as! Int)
+    init(time: NSTimeInterval, icon: String, temperatureMax: Int, temperatureMin: Int) {
+        self.date = NSDate(timeIntervalSince1970: time)
+        self.conditionsDescription = icon
+        self.highTemperature = Temperature(fahrenheitValue: temperatureMax)
+        self.lowTemperature = Temperature(fahrenheitValue: temperatureMin)
     }
+
+    static func forecastFromJSON(JSON: [String: AnyObject]) -> Forecastable {
+        if let time = JSON["time"] as? NSTimeInterval, icon = JSON["icon"] as? String, temperatureMax = JSON["temperatureMax"] as? Int, temperatureMin = JSON["temperatureMin"] as? Int {
+            return DailyForecast(time: time, icon: icon, temperatureMax: temperatureMax, temperatureMin: temperatureMin)
+        } else {
+            return NullDailyForecast()
+        }
+    }
+
+}
+
+@objc class NullDailyForecast: Forecastable {
+    let date: NSDate? = .None
+    let conditionsDescription: String? = .None
+    let highTemperature: Temperature? = .None
+    let lowTemperature: Temperature? = .None
 }
diff --git a/Tropos/Models/TRWeatherUpdate.m b/Tropos/Models/TRWeatherUpdate.m
index 56ce074..1cb24f6 100644
--- a/Tropos/Models/TRWeatherUpdate.m
+++ b/Tropos/Models/TRWeatherUpdate.m
@@ -53,7 +53,7 @@
     NSMutableArray *dailyForecasts = [NSMutableArray array];
 
     for (NSUInteger index = 1; index < 4; index++) {
-        DailyForecast *dailyForecast = [[DailyForecast alloc] initWithJSON:currentConditionsJSON[@"daily"][@"data"][index]];
+        DailyForecast *dailyForecast = [DailyForecast forecastFromJSON:currentConditionsJSON[@"daily"][@"data"][index]];
         [dailyForecasts addObject:dailyForecast];
     }
 
diff --git a/Tropos/ViewModels/TRDailyForecastViewModel.m b/Tropos/ViewModels/TRDailyForecastViewModel.m
index bb068a7..39f6ed5 100644
--- a/Tropos/ViewModels/TRDailyForecastViewModel.m
+++ b/Tropos/ViewModels/TRDailyForecastViewModel.m
@@ -30,19 +30,31 @@
     return [dateFormatter stringFromDate:self.dailyForecast.date];
 }
 
-- (UIImage *)conditionsImage
+- (nullable UIImage *)conditionsImage
 {
-    return [UIImage imageNamed:self.dailyForecast.conditionsDescription];
+    NSString *conditionsDescription = [NSString stringWithFormat:@"%@", self.dailyForecast.conditionsDescription];
+    return [UIImage imageNamed:conditionsDescription];
 }
 
-- (NSString *)highTemperature
+- (nullable NSString *)highTemperature
 {
-    return [[TRTemperatureFormatter new] stringFromTemperature:self.dailyForecast.highTemperature];
+    return [self stringFromTemperature:self.dailyForecast.highTemperature];
 }
 
-- (NSString *)lowTemperature
+- (nullable NSString *)lowTemperature
 {
-    return [[TRTemperatureFormatter new] stringFromTemperature:self.dailyForecast.lowTemperature];
+    return [self stringFromTemperature:self.dailyForecast.lowTemperature];
+}
+
+#pragma mark - Private
+
+- (nullable NSString *)stringFromTemperature:(nullable Temperature *)temperature
+{
+    if (temperature == nil) {
+        return @"";
+    } else {
+        return [[TRTemperatureFormatter new] stringFromTemperature:self.dailyForecast.lowTemperature];
+    }
 }
 
 @end
diff --git a/UnitTests/Tests/DailyForecastSpec.swift b/UnitTests/Tests/DailyForecastSpec.swift
index 3c531e5..6451ac7 100644
--- a/UnitTests/Tests/DailyForecastSpec.swift
+++ b/UnitTests/Tests/DailyForecastSpec.swift
@@ -7,12 +7,12 @@ class DailyForecastSpec: QuickSpec {
         describe("DailyForecast") {
             context("should correctly initialize set properties") {
                 let json: Dictionary<String, AnyObject> = ["time": Double(20.0), "icon": "text", "temperatureMax": 50, "temperatureMin": 10]
-                let dailyForecast = DailyForecast(JSON: json)
-                
-                expect(dailyForecast.date).to(equal(NSDate(timeIntervalSince1970: 20)))
-                expect(dailyForecast.conditionsDescription).to(equal("text"))
-                expect(dailyForecast.highTemperature.fahrenheitValue).to(equal(50))
-                expect(dailyForecast.lowTemperature.fahrenheitValue).to(equal(10))
+//                let dailyForecast = DailyForecast(JSON: json)
+//                
+//                expect(dailyForecast.date).to(equal(NSDate(timeIntervalSince1970: 20)))
+//                expect(dailyForecast.conditionsDescription).to(equal("text"))
+//                expect(dailyForecast.highTemperature.fahrenheitValue).to(equal(50))
+//                expect(dailyForecast.lowTemperature.fahrenheitValue).to(equal(10))
             }
         }
     }
```

tomkowz: Thanks for the update. I started thinking about proper solution and I see we started thinking in similar way. 

1. Personally I don't like passing this "failable" kind of daily forecast just to indicate there is nothing to show. 
2. I like this failable initializer you proposed with class function. Let's take a look on it.

First things first. How view controller works and what RAC is doing:

View controller listen what RAC sends to it and it just iterates through its `TRDailyForecastView` views and update them. We can improve it and do this:

1. Don't create this failable class initializer which complicate things.
2. Don't create failable class which indicates that some daily forecast is not correct (we want to keep it simple so playing with additional object which is empty will complex it).
3. Implement `DailyForecast` class function that validates passed JSON and return proper object.
4. Use your `DailyForecast` initializer which looks good. We'll be always sure that object will be created because JSON validation is elswhere - simplification, Yay!

Conclusion:
- When creating `WeatherUpdate` object we're iterating from 1 to 4 and creating 3 daily forecasts. Let's change it to enumerator that will interate max 3 times but will create up to 3 objects. If something gone wrong it will end with 2 or 1 daily forecast.
- view controller will be notified via RAC that there are some forecasts to display. If there is less than 3 forecasts just hide elements starting from the last. 2 forecasts - show first 2, 1 - show one.

Solution:

    @objc class DailyForecast: NSObject {
        let date: NSDate
        let conditionsDescription: String
        let highTemperature: Temperature
        let lowTemperature: Temperature
        
        init(time: NSTimeInterval, conditionsDescription: String, highTemperature: Int, lowTemperature: Int) {
            self.date = NSDate(timeIntervalSince1970: time)
            self.conditionsDescription = conditionsDescription
            self.highTemperature = Temperature(fahrenheitValue: highTemperature)
            self.lowTemperature = Temperature(fahrenheitValue: lowTemperature)
        }
        
        class func dailyForecastWithJSON(JSON: Dictionary<String, AnyObject>) -> DailyForecast? {
            if let time = JSON["time"] as? NSTimeInterval,
                let conditionsDesc = JSON["icon"] as? String,
                let tempMax = JSON["temperatureMax"] as? Int,
                let tempMin = JSON["temperatureMin"] as? Int {
                return DailyForecast(time: time, conditionsDescription: conditionsDesc, highTemperature: tempMax, lowTemperature: tempMin)
            }
            return nil
        }
    }

I hope I didn't miss anything.

jakecraige: I'll have to think this over a bit. I'd love to get some feedback from @thoughtbot/ios and/or @hyperspacemark on this as well.

tomkowz: Okay, good, let's wait for their opinions.

Do you @jakecraige or someone else will be able to do some review through the weekend? : >

gfontenot: I'll have to read through this thread completely to catch up, but the gist of my thoughts are that we should be using [`Argo`](https://github.com/thoughtbot/Argo) for JSON parsing.

UnitTests/Tests/TRWeatherUpdateCacheSpec.m
```diff
@@ -29,6 +29,14 @@ - (NSURL *)latestWeatherUpdateURL;
     [[NSFileManager defaultManager] removeItemAtURL:weatherUpdateURLForTesting() error:nil];
 };
 
+NSDictionary* (^weatherConditionsWithTemperature) (NSNumber*) = ^NSDictionary* (NSNumber *temp) {
```
gfontenot: Can we put spaces before the pointers? `NSDictionary *`, `NSNumber *`, etc.

tomkowz: I would like not doing this. `TRWeatherUpdateCacheSpec` will be removed in next pull request and replaced with `WeatherUpdateCacheSpec` written in Swift.

gfontenot: Gotcha. I'd rather that we maintain a clean code base wherever possible. That means, in this case, keeping the existing code style for new code even if it's going to be replaced soon.

Tropos/Models/DailyForecast.swift
tomkowz: Didn't use Argo before, Looks nice. I think we can consider using Argo when we'll have more than one thing to decode. Now this is only `DailyForecast` which has very simple structure and IMO integrating additional tool to decode it is a little bit overkill :) What you think?

UnitTests/Tests/TRWeatherUpdateCacheSpec.m
hyperspacemark: > I would like not doing this.

And I'd prefer that we didn't merge code that doesn't conform to the style guide, even if it will be removed eventually.

jakecraige: ^ said style guide here: https://github.com/thoughtbot/guides/tree/master/style/objective_c

I don't think it directly mentions the pointers but the sample has them.

tomkowz: Okay, thanks for the guide. I'll improve the code then.

tomkowz: IMO this should looks like this:

    NSDictionary* (^weatherConditionsWithTemperature) (NSNumber *) = ^NSDictionary* (NSNumber *temp) {

Do you agree? I don't think that asterisk should be placed close to the opening bracket.

Tropos/Models/DailyForecast.swift
tomkowz: I'll update this topic and show how I see implementation of what we already talked about.

`DailyForecast`

	@objc class DailyForecast: NSObject {
	    let date: NSDate
	    let conditionsDescription: String
	    let highTemperature: Temperature
	    let lowTemperature: Temperature
	    
	    init(time: NSTimeInterval, conditionsDescription: String, highTemperature: Int, lowTemperature: Int) {
	        self.date = NSDate(timeIntervalSince1970: time)
	        self.conditionsDescription = conditionsDescription
	        self.highTemperature = Temperature(fahrenheitValue: highTemperature)
	        self.lowTemperature = Temperature(fahrenheitValue: lowTemperature)
	    }
	    
	    class func dailyForecastWithJSON(JSON: Dictionary<String, AnyObject>) -> DailyForecast? {
	        if let time = JSON["time"] as? NSTimeInterval,
	            let conditionsDesc = JSON["icon"] as? String,
	            let tempMax = JSON["temperatureMax"] as? Int,
	            let tempMin = JSON["temperatureMin"] as? Int {
	            return DailyForecast(time: time, conditionsDescription: conditionsDesc, highTemperature: tempMax, lowTemperature: tempMin)
	        }
	        return nil
	    }
	}



`TRWeatherUpdate` - enumerating through forecasts:

    NSMutableArray *dailyForecasts = [NSMutableArray array];
    for (NSUInteger index = 1; index < 4; index++) {
        DailyForecast *dailyForecast = [DailyForecast dailyForecastWithJSON:currentConditionsJSON[@"daily"][@"data"][index]];
        if (dailyForecast != nil) {
            [dailyForecasts addObject:dailyForecast];
        }
    }



`TRWeatherViewController` - updating forecasts views:

    [self.controller.dailyForecastViewModels subscribeNext:^(NSArray *viewModels) {
        NSUInteger index = 0;
        for (TRDailyForecastView *view in forecastViews) {
            TRDailyForecastViewModel *model = index < viewModels.count ? viewModels[index] : nil;
            if (model != nil) {
                view.hidden = NO;
                view.viewModel = model;
            } else {
                view.hidden = YES;
            }
            index++;
        }
    }];

This allows us to hide forecasts that has not been created because of some backend issues. Behaviour the same like proposed by @jakecraige  - https://cloud.githubusercontent.com/assets/3799709/8501575/b69945b6-215b-11e5-8ba3-01f825992b2b.png

jakecraige: As @gfontenot mentioned it would be good to use Argo for the parsing which gives us a similar behavior for free. Since Argo *currently* relies on currying, I wonder if that's even an option right now due to the interop story?

For @tomkowz, what Argo would offer is that `dailyForecastWithJSON` method for free. And it would set us up in the future to use it for more of our JSON parsing as we need it.

I'm not sold on the hiding/showing the dailyForecasts, using the null-object pattern is something we do a lot because it removes logic like the one you defined above. Can you elaborate on why you don't like null-object?

tomkowz: Okay, let's use Argo then.

Hmm, I don't see any advantages of using `NullDailyForecast` - maybe because I never used this pattern. 

Instead of creating nothing we're creating null object - this object means that something goes wrong. 

What's next? Do we create `TRDailyForecastViewModel` with this null object and assign this model to `TRDailyForecastView.model`? How the view handle this? It will change its state internally? Meaning, does this view hide all subviews or itself?

This null object for know is for me the same like:

    @objc class DailyForecast: NSObject {
        let date: NSDate?
        let conditionsDescription: String?
        let highTemperature: Temperature?
        let lowTemperature: Temperature?
        
        lazy var isValid: Bool = {
            return self.date != nil && self.conditionsDescription != nil && self.highTemperature != nil && self.lowTemperature != nil
        }()
    ....

tomkowz: I tried to use Argo but we can't do it accordingly with this answer: http://stackoverflow.com/a/28567439/1046965 - the `decode` method is not visible in Objective-C.

tomkowz: Hey @jakecraige, any updates? :)

UnitTests/Tests/TRWeatherUpdateCacheSpec.m
gfontenot: Hey, sorry, there should always be a space before the pointer, never one after. If we didn't do that elsewhere it slipped through code review. I'd write that line as 

```objc
NSDictionary *(^weatherConditionsWithTemperature) (NSNumber *) = ^NSDictionary *(NSNumber *temp) {
```

tomkowz: Okay, improved.

Tropos/Models/DailyForecast.swift
tomkowz: Hey @jakecraige and @gfontenot - is that possible to discuss this topic more in real time so we can decide about preferred solution faster? Some Slack, Skype, Hangouts, Messages or something?

gfontenot: Sorry we're not more responsive. I appreciate you putting so much effort into this, and feel bad that we're blocking you. We don't actually work on this Mon - Thurs (we're on client work), so during the week we're limited to jumping in on our free time. We should be more responsive on Friday, which is when we have time to do non-client work.

tomkowz: Ahh, okay I see. Let's wait then for you guys.

Tropos/Models/DailyForecast.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+@objc class DailyForecast: NSObject {
```
gfontenot: Is it possible to mark this as `final`? Or is that disallowed by the `@objc` declaration?

tomkowz: Yes, it is. Marked.

gfontenot: Great, I legitimately wasn't sure. We should do this going forward. Making sure we're not subclassing will make the transition to structs easier later.

tomkowz: Sure, will do, this will also make app faster by reducing dynamic dispatch.

Tropos/Models/DailyForecast.swift
tomkowz: Any updates? :)

PR #151, opened 2015-07-16
Use CocoaPods-Keys to hide api keys from git

MarcoSero:
[As suggested](https://github.com/thoughtbot/Tropos/issues/125) by @ashfurrow


MarcoSero:
I had this uncommitted locally for a while, but I just noticed @ashfurrow already proposed the PR. Never mind :+1:

PR #153, opened 2015-08-02
Swedish localization

aenglund:
Added Swedish localization :)

### DIFF
--- a/Resources/Other-Sources/sv.lproj/Localizable.strings
+++ b/Resources/Other-Sources/sv.lproj/Localizable.strings
@@ -1,0 +1,44 @@
+"DifferentTemperatureFormat" = "It's %@ %@ than %@%@.";
+"PrecipitationChanceFormat" = "%@ chance of %@ %@.";
+"SameTemperatureFormat" = "It's the %@ %@ as %@%@.";
+
+"Hotter" = "hotter";
+"Warmer" = "warmer";
+"Cooler" = "cooler";
+"Colder" = "colder";
+"Same" = "same";
+
+"Today" = "today";
+"Yesterday" = "yesterday";
+"Tonight" = "tonight";
+"LastNight" = "last night";
+"ThisMorning" = "this morning";
+"YesterdayMorning" = "yesterday morning";
+"ThisAfternoon" = "this afternoon";
+"YesterdayAfternoon" = "yesterday afternoon";
+
+"NoneRain" = "";
+"SlightRain" = ", with a slight chance of rain";
+"GoodRain" = ", with a good chance of rain";
+"NoneSnow" = ", with no chance of snow";
+"SlightSnow" = ", with a slight chance of snow";
+"GoodSnow" = ", with a good chance of snow";
+
+"North" = "North";
+"Northeast" = "Northeast";
+"East" = "East";
+"Southeast" = "Southeast";
+"South" = "South";
+"Southwest" = "Southwest";
+"West" = "West";
+"Northwest" = "Northwest";
+
+"Good" = "Good";
+"Slight" = "Slight";
+"None" = "None";
+
+"UpdatedAtDateAndTime" = "Updated %@ at %@";
+"UpdatedAtTime" = "Updated at %@";
+
+"CheckingWeather" = "Checking Weather...";
+"UpdateFailed" = "Update Failed";
--- a/Resources/Other-Sources/sv.lproj/Localizable.strings
+++ b/Resources/Other-Sources/sv.lproj/Localizable.strings
@@ -14,7 +14,7 @@
 "Tonight" = "ikväll";
 "LastNight" = "igår kväll";
 "ThisMorning" = "denna morgon";
-"YesterdayMorning" = "igår morgon";
+"YesterdayMorning" = "igår morse";
 "ThisAfternoon" = "denna eftermiddag";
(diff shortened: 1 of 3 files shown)


Resources/Other-Sources/sv.lproj/Localizable.strings
```diff
@@ -0,0 +1,46 @@
+"DifferentTemperatureFormat" = "Det är %@ %@ än %@%@.";
+"PrecipitationChanceFormat" = "%@ chans för %@ %@.";
+"SameTemperatureFormat" = "Det är %@ %@ som %@%@.";
+
+"Hotter" = "hetare";
+"Warmer" = "varmare";
+"Cooler" = "svalare";
+"Colder" = "kallare";
+/* The word temperature is added at the end since the word is needed to make the sentence understandable in this context */
+"Same" = "samma temperatur";
```
mike-burns: Is there something we can do better in the code so that we don't need this workaround? cc @thoughtbot/ios.

Resources/Other-Sources/sv.lproj/Localizable.strings
```diff
@@ -0,0 +1,46 @@
...
+"Northeast" = "Nordöst";
+"East" = "Öst";
+"Southeast" = "Sydöst";
+"South" = "Syd";
+"Southwest" = "Sydväst";
+"West" = "Väst";
+"Northwest" = "Nordväst";
+
+/* These keys are translated so that they can be used in the context of weather prognosis followed by the word for "chance" - "chans" (SWE). */ 
+"Good" = "Stor";
```
mike-burns: This seems especially different. Should the "Good" be changed elsewhere, too?

mike-burns:
This reads fine to me but: ping @thoughtbot/translators .

Resources/Other-Sources/sv.lproj/Localizable.strings
aenglund: I believe that the english sentence "suffers" from the same problem.
Eg. "It's the same today as yesterday" does not state what is the same as yesterday.
The localization I provided would translate to "It's the same temperature today as yesterday" in Swedish. I could change it to match the english sentence if thats preferable? Or maybe the english sentence should be changed?

Resources/Other-Sources/sv.lproj/Localizable.strings
aenglund: This is a bit tricky to explain. The correct translation of "Good" - "Bra" is not very common together with the word "chance" - "chans". Although - it may not be completely wrong, it's just a bit weird to say it like that. So the word I instead chose to translate it to is the swedish word for "Great".

mike-burns: Maybe the translation should be for "GoodChance", then. This is more of a question for @thoughtbot/ios - I'm hoping that this translation file will drive out places where we can improve our translation infrastructure.

Resources/Other-Sources/sv.lproj/Localizable.strings
```diff
@@ -0,0 +1,46 @@
...
+"Colder" = "kallare";
+/* The word temperature is added at the end since the word is needed to make the sentence understandable in this context */
+"Same" = "samma temperatur";
+
+"Today" = "idag";
+"Yesterday" = "igår";
+"Tonight" = "ikväll";
+"LastNight" = "igår kväll";
+"ThisMorning" = "denna morgon";
+"YesterdayMorning" = "igår morgon";
```
calleerlandsson: I'm not sure if this should be `"igår morgon"` or `"igår morse"`. What do you think?

Resources/Other-Sources/sv.lproj/Localizable.strings
```diff
@@ -0,0 +1,46 @@
...
+
+/* These keys are translated so that they can be used in the context of weather prognosis followed by the word for "chance" - "chans" (SWE). */ 
+"Good" = "Stor";
+"Slight" = "Liten";
+"None" = "Ingen";
+
+"UpdatedAtDateAndTime" = "Uppdaterat %@ vid %@";
+"UpdatedAtTime" = "Uppdaterat vid %@";
+
+"CheckingWeather" = "Kontrollerar väder...";
```
calleerlandsson: What do you think of the alternative `"Uppdaterar prognos..."`? Since the error message contains the word "Uppdatering", I thought it might read well.

Resources/Other-Sources/sv.lproj/Localizable.strings
Magnus-G: "igår morse" I think

aenglund: I think `morse` is better as well. Even though its more of a slang word..

Resources/Other-Sources/sv.lproj/Localizable.strings
aenglund: Agree! Fixed that!

jakecraige:
^ @calleerlandsson RR?

calleerlandsson:
Looks good to me.

jakecraige:
Thanks! Merged in https://github.com/thoughtbot/Tropos/commit/dc2ed3d407a6ee3099fbc630a24fb70f43d9fbec

PR #154, opened 2015-08-24
Switch to GNU GPL 3

mike-burns:
This is to reflect our feelings on
https://twitter.com/GFontenot/status/635801565048229890

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,22 +1,676 @@
-Copyright (c) 2014 thoughtbot, inc.
+Tropos is copyright 2014, 2015 thoughtbot, inc.
 
-MIT License
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
 
-Permission is hereby granted, free of charge, to any person obtaining
-a copy of this software and associated documentation files (the
-"Software"), to deal in the Software without restriction, including
-without limitation the rights to use, copy, modify, merge, publish,
-distribute, sublicense, and/or sell copies of the Software, and to
-permit persons to whom the Software is furnished to do so, subject to
-the following conditions:
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
 
-The above copyright notice and this permission notice shall be
-included in all copies or substantial portions of the Software.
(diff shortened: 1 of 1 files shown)


gfontenot:
Can we add more info to the commit about what the ramifications for this change are?

tute:
It looks like we can still be in the App Store, because we hold the copyrights? https://apple.stackexchange.com/questions/6109/is-it-possible-to-have-gpl-software-in-the-mac-app-store

BlakeWilliams:
@tute I think that's correct. I think this would prevent everyone else from being able to publish any derivatives or copies to the app store license wise.

sgrif:
Does this actually match our feelings on the matter? I personally don't think we should disallow all derivations from being sold. For me, having someone use this as a baseline to make something clearly different even if it was based on our work is fine. Someone uploading as-is and calling it theirs is not. Of course that's overly vague, and would never work for a legal document like this.

Would it make sense for us to focus on requiring attribution, instead of disallowing entirely instead?

mike-burns:
The interesting sections are 4, 5, 6, and 10:

- "You may charge any price or no price for each copy that you convey, and you may offer support or warranty protection for a fee."
- If someone changes and re-distributes the source code, it must note that it is a modification and it cannot change the license or copyright.
- We must provide access to the source code to anyone who has the binary; this requirement applies to anyone else who distributes the binary, even if they modify the code.

This differs both in law and in spirit from the MIT license we currently have: MIT allows us to charge any price (just like GNU GPL) but does not require the other two bullet points -- especially for people who fork. If someone forks a MIT project, they can essentially do what they want without any mention of us.

mike-burns:
Do we want this still, @gfontenot @tonyd256 ?

gfontenot:
I forgot this was a thing, so probably not?

jakecraige:
Perhaps we should reevaluate, I just found another blatant clone... What Weather by Ross Tsingos
https://appsto.re/us/SpPU-.i

tonyd256:
The license will only be useful if we decide to pursue violators. Is this something we care about doing/is worth our time? Or do you think people will be better open source citizens just by having this license?

mike-burns:
Currently, no one is violating the Tropos license. Switching to GNU GPL would make the new code protected from this kind of behavior, and allow us to pursue those violations if we want.

PR #158, opened 2015-09-18
Fix bug incorrectly displaying the chance of rain

Euraldius:
forecast.io gives us the percentage chance of rain from 0...1, not 0...100 (as
our code previously assumed)

* Also renames the `todaysConditions` variable to `currentConditions` for clarity

Trello: https://trello.com/c/N3qcunas/35-reports-100-precipitation-as-sight-chance-of-rain

Pair: Adam Sharp

### DIFF
--- a/Tropos/Models/Precipitation.swift
+++ b/Tropos/Models/Precipitation.swift
@@ -10,7 +10,7 @@ import Foundation
 
     var chance: PrecipitationChance {
         switch probability {
-        case _ where probability > 30.0: return .Good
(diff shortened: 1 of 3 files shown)


jakecraige:
Oh no! Good catch. :+1: 

Was wondering why I never saw good chance of rain while in NYC, hah

hyperspacemark:
Nice fix! I actually noticed this behavior last night. I'm in Detroit and it was _pouring rain for hours_ and Tropos was saying there was a slight chance of rain tonight. :+1:

PR #159, opened 2015-09-26
Update project for Xcode 7 and Swift 2

jakecraige:
- Required config updates
- Remove redundant Printable protocol
- Fix custom Temperature compare function args
- Move storyboard from Sources into Resources
  More info: https://github.com/thoughtbot/liftoff/issues/239
- Fix naming conflict of `describe` from Quick and Specta
  in TRTemperatureComparisonFormatterSpec.m by using Quick instead
  of Specta.
- Update Pod dependencies
- Update Carthage dependencies

### DIFF
--- a/Tropos/Models/Temperature.swift
+++ b/Tropos/Models/Temperature.swift
@@ -13,7 +13,7 @@ func -(lhs: Temperature, rhs: Temperature) -> Temperature {
     return Temperature(fahrenheitValue: lhs.fahrenheitValue - rhs.fahrenheitValue)
 }
 
-@objc class Temperature: NSObject, Printable {
+@objc class Temperature: NSObject {
     let fahrenheitValue: Int
     
     lazy var celsiusValue: Int = {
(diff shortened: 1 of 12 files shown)


Tropos/Models/Temperature.swift
```diff
@@ -13,7 +13,7 @@ func -(lhs: Temperature, rhs: Temperature) -> Temperature {
     return Temperature(fahrenheitValue: lhs.fahrenheitValue - rhs.fahrenheitValue)
 }
 
-@objc class Temperature: NSObject, Printable {
+@objc class Temperature: NSObject {
```
hyperspacemark: Couldn't we still declare conformance to `CustomStringConvertible`? That would let us string-interpolate an instance of `Temperature` in the future.

hyperspacemark:
:+1:

Tropos/Models/Temperature.swift
jakecraige: If I remember correctly, I think Xcode was now complaining that it was redundant since NSObject already does it. We still have the implementation https://github.com/thoughtbot/Tropos/blob/jc-swift-upgrade/Tropos/Models/Temperature.swift#L42-L45

hyperspacemark: Aha! Didn't know that. Cool. Thanks!

PR #160, opened 2015-10-16
Bump version for 1.1.1

jakecraige:
Had to disable bitcode because of some dependencies


gfontenot:
Why are we disabling bit code?

jakecraige:
I was being lazy when I first set this up... I needed to update a few dependencies. It was pretty simple to do. Updated.

gfontenot:
:+1:

PR #162, opened 2015-11-30
Update CocoaPods gem to match Podfile.lock 

### LINKED ISSUE #161: mismatched CocoaPods version Gemfile.lock vs Podfile.lock
Podfile.lock created with version 0.38.2: https://github.com/thoughtbot/Tropos/blob/master/Podfile.lock#L48
Gemfile.lock has cocoapods versioned at 0.37.2
https://github.com/thoughtbot/Tropos/blob/master/Gemfile.lock#L11

Resulting in this warning in `bin/setup`:

```
[!] The version of CocoaPods used to generate the lockfile (0.38.2) is higher than the version of the current executable (0.37.2). Incompatibility issues may arise.
```

and possibly related error later in `bin/setup` : `Generating Pods project`
screen shot 2015-11-30 at 1 59 32 pm

And, lastly, in workspace:

screen shot 2015-11-30 at 2 02 54 pm

Seeing if I can whip up a quick fix.

sebbean:
Closes #161

Should we add `Carthage/` to gitignore? Just hangs out in stage otherwise.


Tropos.xcodeproj/project.pbxproj
```diff
@@ -830,6 +832,36 @@
 			shellPath = /bin/sh;
 			shellScript = "/usr/local/bin/carthage copy-frameworks";
 		};
+		E9D11CFCA967ADDF3B704D82 /* Embed Pods Frameworks */ = {
```
jakecraige: Is this a new thing required by CocoaPods 0.39?

Gemfile.lock
```diff
@@ -55,6 +60,3 @@ PLATFORMS
 DEPENDENCIES
   cocoapods
   xcpretty
-
-BUNDLED WITH
```
jakecraige: If you upgrade your `bundler` version this should come back. I'd prefer to still have it here since others running it will generate this part.

jakecraige:
Thanks!

> Should we add `Carthage/` to gitignore? Just hangs out in stage otherwise.

Yeah it definitely should.

Tropos.xcodeproj/project.pbxproj
sebbean: Yea, it showed up after `pod install`

PR #166, opened 2016-04-12
Update dependencies to fix bitcode linker errors when building for device

sharplet:
This change updates both Mixpanel and HockeySDK to new minor versions that include support for bitcode.

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -1,8 +1,8 @@
 platform :ios, '8.0'
 
-pod 'HockeySDK', '~> 3.6.2', :inhibit_warnings => true
+pod 'HockeySDK', '~> 3.6', :inhibit_warnings => true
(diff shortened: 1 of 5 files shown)


sharplet:
This build failure was hidden because CI doesn't currently build for device. I'm going to look into setting up a device build as part of this PR.

gfontenot:
👍 Should be possible to build for device by using `platform=iOS/Generic` in the build script. Maybe have it build and run the tests, then do a second build for device?

sharplet:
I've pushed a new `bin/archive` script that ultimately performs an `xcodebuild archive`. The `archive` action builds for device, and I think having a successful `archive` step is a stronger signal from CI than a `build` that targets the generic device destination. Let me know what you think.

gfontenot:
makes sense to me 👍

PR #167, opened 2016-04-18
Add tests for TRSettingsController

sharplet:
This is the first in a series of PRs to facilitate porting more files to Swift. A future PR will actually port TRSettingsController to Swift; this PR is intended to add test coverage.

While adding tests I also ran into some issues with build settings related to the PCH file. This PR also removes PCH files in favour of explicit imports.


Tropos/Controllers/TRSettingsController.m
```diff
@@ -2,38 +2,32 @@
...
-#pragma mark - Properties
-
-- (RACSignal *)unitSystemChanged
+- (instancetype)init
 {
-    return [[[self.userDefaultsChanged filter:^BOOL(NSNotification *notification) {
-        return [notification.name isEqualToString:NSUserDefaultsDidChangeNotification];
-    }] map:^id(id value) {
-        return @([self unitSystem]);
-    }] startWith:@([self unitSystem])];
```
sharplet: This method has moved [here](https://github.com/thoughtbot/Tropos/pull/167/files#diff-50402482dc3839f4d529b6e1a18d494eR7).

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -44,4 +49,10 @@ - (void)application:(UIApplication *)application performFetchWithCompletionHandl
     }];
 }
 
+- (BOOL)isCurrentlyTesting
+{
+    NSString *bundleInjectionPath = [[[NSProcessInfo processInfo] environment] objectForKey:@"XCInjectBundleInto"];
```
dazmuda: Could you maybe use a run argument or preproc macro instead?

Tropos/Controllers/TRWeatherController.m
```diff
@@ -37,6 +39,7 @@ - (instancetype)init
     self.geocodeController = [TRGeocodeController new];
     self.forecastController = [TRForecastController new];
     self.settingsController = [TRSettingsController new];
+    self.unitSystemChanged = [[self.settingsController unitSystemChanged] replayLastLazily];
```
dazmuda: Does this need to be lazy? As in, is there a reason we wait for a subscription to open the multicast connection or is just for cleanliness?

dazmuda:
Small questions. Looks good.

Tropos/Controllers/TRAppDelegate.m
sharplet: Good idea, check out 832f613

Tropos/Controllers/TRWeatherController.m
sharplet: It certainly wouldn't break without it.

Mainly: I'm not aware of any downside to the laziness, and I think deferring the work until it's actually needed is generally a good idea.

sharplet:
Thanks @dazmuda ✨

PR #168, opened 2016-04-18
Port remaining models to Swift

sharplet:
In porting `WeatherUpdate` and `WeatherUpdateCache`, it became clear that `SettingsController` also need to be ported, so things would still work nicely together. I've also converted `TemperatureComparisonFormatter` in the process.

The other main change I made was to reintroduce the `TR` prefix for the objc names of each model. Even though Swift has module namespacing, objc types still live in the global namespace.


Tropos/Controllers/SettingsController.swift
```diff
@@ -0,0 +1,40 @@
...
+        set {
+            NSUserDefaults.standardUserDefaults().setInteger(newValue.rawValue, forKey: TRSettingsUnitSystemKey)
+        }
+    }
+
+    init(locale: NSLocale) {
+        self.locale = locale
+    }
+
+    convenience override init() {
```
klaaspieter: Have you considered using `init(locale: NSLocale = .autoupdatingCurrentLocale())` instead of a convenience initializer?

Tropos/Controllers/SettingsController.swift
```diff
@@ -0,0 +1,40 @@
...
+        self.init(locale: .autoupdatingCurrentLocale())
+    }
+
+    func registerSettings() {
+        registerUnitSystem()
+        registerLastVersion()
```
klaaspieter: The `synchronize` call to `NSUserDefaults` that is in Objective-C is missing here.

Tropos/Formatters/TemperatureComparisonFormatter.swift
```diff
@@ -0,0 +1,27 @@
...
+            ? NSLocalizedString("SameTemperatureFormat", comment: "")
+            : NSLocalizedString("DifferentTemperatureFormat", comment: "")
+
+        return (
+            description: String(format: format, adjective, timeOfDay, timeOfYesterday, precipitation),
+            adjective: adjective
+        )
+    }
+
+    @available(*, unavailable, message="use localizedStrings(fromComparison:precipitation:date:) instead")
```
klaaspieter: Have you considered replacing all usages of this method with the new method?

Tropos/Formatters/TimeOfDay.swift
```diff
@@ -0,0 +1,52 @@
...
+
+extension NSCalendar {
+    func timeOfDay(forDate date: NSDate) -> TimeOfDay {
+        let components = self.components(.Hour, fromDate: date)
+
+        if components.hour < 4 {
+            return .Night
+        } else if components.hour < 9 {
+            return .Morning
+        } else if components.hour < 14 {
```
klaaspieter: Isn't everything after noon “afternoon”? Isn't this also missing evening (18:00-00:00)?

Tropos/Models/DailyForecast.swift
```diff
@@ -0,0 +1,24 @@
+import Foundation
+
+@objc(TRDailyForecast) class DailyForecast: NSObject {
+    let date: NSDate
+    let conditionsDescription: String
+    let highTemperature: Temperature
+    let lowTemperature: Temperature
+
+    init?(JSON: AnyObject?) {
```
klaaspieter: Probably not applicable to this PR, but do you know the reason we're not using Argo?

klaaspieter: Answering my own question with a semi question: Because it was Objective-C?

Is it worth considering changing this to Argo now that it's Swift?

Tropos/Models/WeatherUpdateCache.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    convenience init(fileName: String) {
+        let fileManager = NSFileManager.defaultManager()
+        guard let cachesURL = fileManager.URLsForDirectory(.CachesDirectory, inDomains: .UserDomainMask).first else {
+            fatalError("Unable to locate user caches directory")
+        }
+        self.init(fileName: fileName, inDirectory: cachesURL)
+    }
+
+    convenience override init() {
```
klaaspieter: Have you considering `init(fileName: String = TRLatestWeatherUpdateFileName)`?

UnitTests/Tests/WeatherUpdateCacheSpec.swift
```diff
@@ -0,0 +1,68 @@
...
+        .URLsForDirectory(.DocumentDirectory, inDomains: .UserDomainMask)
+        .first!
+}
+
+private var testWeatherUpdateURL: NSURL {
+    return testCachesDirectory.URLByAppendingPathComponent(testCacheFileName)
+}
+
+private func resetFilesystem() {
+    do {
```
klaaspieter: Have you considered replacing this with a `try!`. This is cleaning up state, I would want the current test to fail if for some reason the cleanup fails. 

Swallowing the error could mean a later test fails because an earlier test didn't correctly clean up the state.

UnitTests/Tests/WeatherUpdateSpec.swift
```diff
@@ -0,0 +1,61 @@
...
+        "daily": ["data": [data]],
+    ]
+}
+
+final class WeatherUpdateSpec: QuickSpec {
+    override func spec() {
+        describe("TRWeatherUpdate") {
+            context("currentTemp is higher than currentHigh") {
+                it("updates currentHigh to match") {
+                    let conditions = weatherConditions(temperature: 70)
```
klaaspieter: Have you considered placing line breaks between [arrange, act and assert] phase?

[arrange, act and assert]: http://c2.com/cgi/wiki?ArrangeActAssert

I would put line breaks between the three phases of these tests: setup, exercise, and verification. As these tests grow, it will be more important to clearly delineate what's happening where.
 Add a line note

klaaspieter:
I had some questions and minor comments, but LGTM aside from the circle failure.

Tropos/Controllers/SettingsController.swift
sharplet: If I remove the convenience initialiser, I get this runtime crash:

```
Tropos/Controllers/SettingsController.swift: 3: 41: fatal error: use of unimplemented initializer 'init()' for class 'Tropos.SettingsController'
```

It seems that Swift doesn't synthesize the default initialiser for you, and implementing your own designated initialiser removes the inherited `-[NSObject init]`.

Tropos/Controllers/SettingsController.swift
sharplet: Here's a snippet from the docs:

> ... use this method only if you cannot wait for the automatic synchronization (for example, if your application is about to exit) ...

Because this is being called at application startup, the explicit `-synchronize` isn't necessary, which is why I removed it.

Tropos/Formatters/TemperatureComparisonFormatter.swift
sharplet: At this stage, this method is still used from Objective-C. Marking this as unavailable here is just to prevent it being used from Swift.

Tropos/Controllers/SettingsController.swift
klaaspieter: Out of curiosity is the initializer being called from Objective-C? It makes sense that Objective-C wouldn't see the argument's default value.

Tropos/Controllers/SettingsController.swift
klaaspieter: I've had bad experiences with not calling `-synchronize` hence the question. YMMV.

Tropos/Formatters/TemperatureComparisonFormatter.swift
sharplet: A future PR will tackle porting `TRWeatherViewModel` to Swift (and its remaining dependencies), and removing this version of the method.

Tropos/Formatters/TimeOfDay.swift
sharplet: It's possible this could be improved—would you be interested in raising an issue where we can discuss it? For now I believe this faithfully maintains the behaviour of the existing Objective-C version.

Tropos/Models/DailyForecast.swift
sharplet: I think that would potentially be a useful improvement in the future, when more of the app has been ported and we can drop Objective-C compatibility for the model layer. I think it's worth raising an issue to track this.

Tropos/Models/WeatherUpdateCache.swift
sharplet: See my previous answer regarding using these initialisers from Objective-C.

UnitTests/Tests/WeatherUpdateCacheSpec.swift
sharplet: When I was testing this, `removeItemAtURL(_:)` would throw if you tried to remove a path that doesn't exist. So this is essentially `rm -f`, where if the file already doesn't exist it's no problem.

Tropos/Controllers/SettingsController.swift
sharplet: It is—so that's the reason I've written it this way currently. Mix-and-match!

Tropos/Controllers/SettingsController.swift
sharplet: Would you mind elaborating? I seem to remember seeing conversations recently of a change (improvement) in behaviour. Perhaps in a recent version of iOS.

gfontenot: [`synchronize` is basically a synonym for `sleep` at this point](https://twitter.com/catfish_man/status/674727133017587712) and really shouldn't be needed any longer.

gfontenot: More from David Smith on the issue:

https://twitter.com/Catfish_Man/status/674727133017587712
https://twitter.com/Catfish_Man/status/674734312860737536
https://twitter.com/Catfish_Man/status/674728434648543232
https://twitter.com/Catfish_Man/status/674731285156249600

Tropos/Controllers/SettingsController.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+@objc(TRSettingsController) final class SettingsController: NSObject {
```
gfontenot: TIL you could specify a new name for objective c. Neat.

Tropos/Controllers/SettingsController.swift
gfontenot: Can we add this as a default value to the `init(locale:)` method so that we can _just_ remove this when we stop calling it from objective c?

gfontenot: Oh, or maybe that would result in ambiguous method declarations.

Tropos/Controllers/SettingsController.swift
```diff
@@ -0,0 +1,40 @@
...
+    private func registerUnitSystem() {
+        let localeUsesMetric = locale.objectForKey(NSLocaleUsesMetricSystem)?.boolValue ?? false
+        let unitSystem: UnitSystem = localeUsesMetric ? .Metric : .Imperial
+        NSUserDefaults.standardUserDefaults().registerDefaults([TRSettingsUnitSystemKey: unitSystem.rawValue])
+    }
+
+    private func registerLastVersion() {
+        let version = NSBundle.mainBundle().versionNumber
+        NSUserDefaults.standardUserDefaults().registerDefaults([TRSettingsLastVersionKey: version])
+        NSUserDefaults.standardUserDefaults().setObject(version, forKey: TRSettingsLastVersionKey)
```
gfontenot: Do we need both of these? I'm unclear on why the first would ever be needed if we're always doing the second.

Tropos/Formatters/TRBearingFormatter.h
```diff
@@ -1,6 +1,6 @@
 @interface TRBearingFormatter : NSObject
 
-+ (NSString *)cardinalDirectionStringFromBearing:(CGFloat)bearing;
-+ (NSString *)abbreviatedCardinalDirectionStringFromBearing:(CGFloat)bearing;
++ (NSString *)cardinalDirectionStringFromBearing:(double)bearing;
++ (NSString *)abbreviatedCardinalDirectionStringFromBearing:(double)bearing;
```
gfontenot: What's this about?

sharplet: `CGFloat` is `double` on 64-bit architectures, and `float` on 32-bit architectures. When compiling for 32-bit, passing a `double` to this method generated a loss of precision warning. When archiving, this warning was treated as an error and failed the build.

sharplet: On top of that, using `CGFloat` in the model layer here adds a dependency on `QuartzCore` above just `Foundation`, which is unnecessary.

Tropos/Controllers/SettingsController.swift
sharplet: I think it would.

gfontenot:
👍

Tropos/Controllers/SettingsController.swift
sharplet: I did consider removing this too. I think I might.

UnitTests/Tests/WeatherUpdateSpec.swift
sharplet: I think I must have an internal line number threshold for determining whether to insert extra line breaks in these situations. It appears to be at around 4 lines. 😛

PR #169, opened 2016-04-20
Port view models to Swift

sharplet:
I've ported all the view models, and the formatters they depend on, to Swift. I was also able to remove the last traces of Specta & Expecta in the process.

I've also ported SettingsController, as the view models depend on the current unit system setting in NSUserDefaults.


Tropos/Formatters/TemperatureComparisonFormatter.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
-@objc(TRTemperatureComparisonFormatter) class TemperatureComparisonFormatter: NSObject {
-    static func localizedStrings(fromComparison comparison: TemperatureComparison, precipitation: String, date: NSDate) -> (description: String, adjective: String) {
+struct TemperatureComparisonFormatter {
```
gfontenot: yessss

Tropos/Models/CardinalDirection.swift
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    var description: String {
+        return rawValue
+    }
+
+    var abbreviation: String {
+        let characters = "NSEW".characters
+        return String(description.characters.filter { characters.contains($0) })
```
gfontenot: this is neat

Tropos/Models/DailyForecast.swift
```diff
@@ -1,10 +1,10 @@
...
-@objc(TRDailyForecast) class DailyForecast: NSObject {
-    let date: NSDate
-    let conditionsDescription: String
-    let highTemperature: Temperature
-    let lowTemperature: Temperature
+struct DailyForecast {
+    var date: NSDate
+    var conditionsDescription: String
+    var highTemperature: Temperature
+    var lowTemperature: Temperature
```
gfontenot: Why did these need to become mutable?

UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
```diff
@@ -19,7 +19,7 @@ @interface TRWeatherController (Tests)
 
 @end
 
-SpecBegin(TRApplicationController)
+QuickSpecBegin(TRApplicationControllerSpec)
```
gfontenot: nice

UnitTests/Tests/Formatters/TemperatureFormatterSpec.swift
```diff
@@ -0,0 +1,26 @@
...
+            expect(formatter.stringFromTemperature(temperature)) == "0°"
+        }
+
+        it("can format celsius temperatures in fahrenheit") {
+            let formatter = TemperatureFormatter(unitSystem: .Imperial)
+            let temperature = Temperature(celsiusValue: 0)
+            expect(formatter.stringFromTemperature(temperature)) == "32°"
+        }
+
+        it("uses the default unit system in settings if not specified") {
```
gfontenot: I don't see how this test exercises this description

gfontenot:
A few minor questions, but this looks great to me 👍

Tropos/Models/DailyForecast.swift
sharplet: Main reason is it's more flexible, and allows you to use the type with `var` in the expected way. It's still safe because of value semantics.

UnitTests/Tests/Formatters/TemperatureFormatterSpec.swift
sharplet: `TemperatureFormatter.init()` uses a default `SettingsController()` internally to determine the unit system, as apposed to `.init(unitSystem:)` which uses the explicitly passed-in unit system. It then results an a celsius-formatted temperature, matching the unit system in settings.

Does that help at all?

sharplet: Would a description like "it fetches the unit system from the SettingsController by default" be an improvement?

gfontenot: Ah, yes. That would make more sense. Seeing "default unit system" followed by explicitly setting a `unitSystem` property was confusing me.

PR #170, opened 2016-04-27
Upgrade cocoapods to 1.0.0 RC 2

klaaspieter:
According to https://github.com/CocoaPods/CocoaPods/issues/4345#issuecomment-147738502 Cocoapods 0.39 is broken with Ruby 2.3.0. Since we're going to have to upgrade at some point anyway, I figured why not now.

This upgrades cocoapods to the latest 1.0.0 beta (8). It also updates the Podfile, which was still using deprecated methods that have been removed in 1.0.0. Finally Acknowledgements are now generated using the `cocoapods-acknowledgements` plugin.


sharplet:
Looks good to me. Would you mind rebasing?

klaaspieter:
@sharplet I moved the Courier dependency to Carthage instead. Only issue I can see is that the Cocoapods acknowledgements won't work anymore using the current setup.

I also updated to using cocoapods rc1, which was just released today, and made the specs pass again.

Finally I added a .ruby-version file and set `2.3.0` in the Gemfile.

PR #171, opened 2016-04-28
Handle content-available notifications

klaaspieter:
The backend sends a content-available notification after which the app
can check the current weather and present the user with a local
notification.

This prevents the backend from having to know about the forecast.io API
and keeps all the relevant logic in the app.

I intend to make the courier gem and ios framework public before merging this.

I also think we should have different notification contents. Suggestions are welcome.

### DIFF
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,2 +1,3 @@
-github "Quick/Nimble" "v4.0.0"
+github "Quick/Nimble" "v4.0.1"
 github "Quick/Quick" "v0.9.2"
+github "thoughtbot/courier-ios" "v0.0.1"
--- a/Podfile
+++ b/Podfile
@@ -1,4 +1,5 @@
 platform :ios, '8.0'
+use_frameworks!
 
 pod 'HockeySDK', '~> 3.6', :inhibit_warnings => true
 pod 'ReactiveCocoa', '~> 2.4.7'
--- a/Tropos.xcodeproj/project.pbxproj
+++ b/Tropos.xcodeproj/project.pbxproj
@@ -1,1258 +1,3819 @@
-// !$*UTF8*$!
-{
-	archiveVersion = 1;
-	classes = {
-	};
-	objectVersion = 46;
-	objects = {
-
-/* Begin PBXBuildFile section */
-		46660A27B04F4C7B78A7920F /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 436D9431497525313B32E225 /* Foundation.framework */; };
-		4A2F34C81CBE808A00417986 /* DailyForecast.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A2F34C71CBE808A00417986 /* DailyForecast.swift */; };
-		4A2F34CC1CBE8DF200417986 /* WeatherUpdate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A2F34CB1CBE8DF200417986 /* WeatherUpdate.swift */; };
-		4A2F34CF1CBE9B4500417986 /* TRWeatherUpdate+Analytics.m in Sources */ = {isa = PBXBuildFile; fileRef = 4DEDF4F51A6FA682006F3F11 /* TRWeatherUpdate+Analytics.m */; };
-		4A2F35111CBEC53800417986 /* New York.placemark in Resources */ = {isa = PBXBuildFile; fileRef = 4A2F35101CBEC53800417986 /* New York.placemark */; };
-		4A2F35131CBEC6D600417986 /* WeatherUpdateSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A2F35121CBEC6D600417986 /* WeatherUpdateSpec.swift */; };
-		4A2F35151CBECBD800417986 /* TemperatureComparisonFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A2F35141CBECBD800417986 /* TemperatureComparisonFormatter.swift */; };
-		4A2F35171CBED29C00417986 /* TimeOfDay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A2F35161CBED29C00417986 /* TimeOfDay.swift */; };
-		4A3F7D6E1CBEDD33005DF19E /* WeatherUpdateCacheSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7E4A3A11B069BB500F489E3 /* WeatherUpdateCacheSpec.swift */; };
-		4A3F7D701CBEDD6E005DF19E /* TestPlacemark.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F7D6F1CBEDD6E005DF19E /* TestPlacemark.swift */; };
-		4A3F7D721CBEDFC5005DF19E /* WeatherUpdateCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4A3F7D711CBEDFC5005DF19E /* WeatherUpdateCache.swift */; };
-		4AC287A41CBC3D800064F48A /* Nimble.framework in Embed Carthage Frameworks */ = {isa = PBXBuildFile; fileRef = 6D0F135A1B432976001685BA /* Nimble.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
-		4AC287A51CBC3D820064F48A /* Quick.framework in Embed Carthage Frameworks */ = {isa = PBXBuildFile; fileRef = 6D0F135B1B432976001685BA /* Quick.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
-		4ACA583B1CC1884400DD0CDC /* SettingsController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4ACA583A1CC1884400DD0CDC /* SettingsController.swift */; };
-		4ACA583D1CC188EB00DD0CDC /* UnitSystem.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4ACA583C1CC188EB00DD0CDC /* UnitSystem.swift */; };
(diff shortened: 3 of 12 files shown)


klaaspieter:
I'll _need_ to make things public for CI to pass.

Podfile
```diff
@@ -1,8 +1,10 @@
 platform :ios, '8.0'
+use_frameworks!
 
 pod 'HockeySDK', '~> 3.6', :inhibit_warnings => true
 pod 'ReactiveCocoa', '~> 2.4.7'
 pod 'Mixpanel', '~> 2.7', :inhibit_warnings => true
+pod 'Courier', git: "https://github.com/thoughtbot/courier-ios.git", branch: "kp-podspec"
```
sharplet: Did you consider bringing this in via Carthage?

Tropos.xcodeproj/project.pbxproj
```diff
@@ -998,7 +1005,6 @@
 				COPY_PHASE_STRIP = YES;
 				DEFINES_MODULE = YES;
 				DSTROOT = /tmp/xcodeproj.dst;
-				EMBEDDED_CONTENT_CONTAINS_SWIFT = YES;
```
sharplet: Why was this setting disabled?

Tropos/TRCourierClient.swift
```diff
@@ -0,0 +1,13 @@
+import Courier
+
+@objc class TRCourierClient: NSObject {
```
sharplet: Elsewhere I've been using the un-prefixed name for Swift, and then only prefixing the Objective-C name. What do you think about this change?

```diff
-@objc class TRCourierClient: NSObject {
+@objc(TRCourierClient) final class CourierController: NSObject {
```

Tropos/TRCourierClient.swift
```diff
@@ -0,0 +1,13 @@
...
+    let instance: Courier
+
+    init(apiToken: String) {
+        instance = Courier(apiKey: apiToken, environment: .Development)
+    }
+
+    func subscribeToChannel(channel: String, withToken token: NSData) {
+        instance.subscribeToChannel(channel, withToken: token)
+    }
+}
```
sharplet: Missing a new line at the end of this file.

Tropos/TRCourierClient.swift
```diff
@@ -0,0 +1,13 @@
+import Courier
+
+@objc class TRCourierClient: NSObject {
+    let instance: Courier
```
sharplet: Should this property be marked `private`?

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -38,6 +44,35 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
     return YES;
 }
 
+- (void)application:(UIApplication *)application didRegisterForRemoteNotificationsWithDeviceToken:(NSData *)deviceToken;
+{
+    NSString *channel = [[NSTimeZone defaultTimeZone] name];
```
sharplet: Given that we're not holding onto this reference, I believe that `+defaultTimeZone` and `+localTimeZone` are basically equivalent in this usage. However, I'm wondering if perhaps `+localTimeZone` would be more appropriate, semantically, for Tropos?

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -38,6 +44,35 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    [self.courier subscribeToChannel:channel withToken:deviceToken];
+}
+
+- (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(nonnull void (^)(UIBackgroundFetchResult))completionHandler;
+{
+    if (![userInfo valueForKeyPath:@"aps.content-available"]) {
```
sharplet: Do you think it's worth checking the actual value here, rather than just the presence of the key? Technically the value should be `@1`.

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -38,6 +44,35 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+        [application scheduleLocalNotification:notification];
+
+        completionHandler(UIBackgroundFetchResultNewData);
+
+    } error:^(NSError *error) {
+        completionHandler(UIBackgroundFetchResultFailed);
+    }];
+
+    [controller.updateWeatherCommand execute:self];
+}
```
sharplet: I'm thinking that the bulk of this logic might be better moved to `TRApplicationController`, to keep the application delegate method small. Similar to the background fetch scenario.

That would also mean that you can reuse the same weather controller instance. The main reason I think this is beneficial is that the internal `RACCommand` inside the weather controller ensures that only one weather update is ever firing at a time. But if you have two separate instances, you can technically have multiple requests running concurrently (e.g., if a push notification is received while the app is already in the foreground).

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -54,4 +89,12 @@ - (BOOL)isCurrentlyTesting
     return [[NSUserDefaults standardUserDefaults] boolForKey:@"TRTesting"];
 }
 
+- (TRCourierClient *)courier;
+{
+    if (!_courier) {
+        _courier = [[TRCourierClient alloc] initWithApiToken: TRCourierAPIToken];
```
sharplet: I think that the application controller should probably hold onto this instance, as that's where the other API integrations live.

sharplet:
This looks great! Can't wait to get this happening.

A few comments, mostly open questions about where certain code should belong.

Tropos.xcodeproj/project.pbxproj
klaaspieter: Because Cocoapods complained about it. I deleted the build setting to set it's back to it's default. Cocoapods will set it to the value it needs using it's xcconfig.

klaaspieter:
I added https://trello.com/c/TkKZfa8t/41-better-permission-priming to discuss better permission priming.

Cartfile.resolved
```diff
@@ -1,2 +1,3 @@
-github "Quick/Nimble" "v4.0.0"
+github "Quick/Nimble" "v4.0.1"
 github "Quick/Quick" "v0.9.2"
+github "thoughtbot/courier-ios" "v0.0.1"
```
sharplet: Can we switch to semantic versions here? For example:

```
github "Quick/Nimble" ~> 4.0
github "Quick/Quick" ~> 0.9
github "thoughtbot/courier-ios" ~> 0.0.1
```

Tropos/Controllers/TRApplicationController.m
```diff
@@ -40,4 +47,12 @@ - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)applica
     }
 }
 
+- (TRCourierClient *)courier;
+{
+    if (!_courier) {
+        _courier = [[TRCourierClient alloc] initWithApiToken: TRCourierAPIToken];
```
sharplet: Was there a specific reason this isn't in `-init`? I think it would be a good idea to follow the existing initialization pattern where appropriate.

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -38,6 +40,32 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    [self.courier subscribeToChannel:channel withToken:deviceToken];
+}
+
+- (void)application:(UIApplication *)application didReceiveRemoteNotification:(NSDictionary *)userInfo fetchCompletionHandler:(nonnull void (^)(UIBackgroundFetchResult))completionHandler;
+{
+    if (![userInfo[@"aps"][@"content-available"] isEqual: @1]) {
+        completionHandler(UIBackgroundFetchResultNoData);
+        return;
+    }
+    [self.applicationController.conditionsDescription subscribeNext:^(NSAttributedString *conditions) {
```
sharplet: This is going to create a new subscription whenever a new notification is received, which could be a potential memory leak. What do you think about chaining a call to `take:1`, or something else similar?

sharplet: Instead of exposing the `conditionsDescription` property on `TRApplicationController`, what are your thoughts on instead exposing a signal that sends `UILocalNotification`s ready for scheduling? That may simplify the responsibilities of the app delegate, mainly keeping the "view" responsibilities elsewhere.

Tropos/Controllers/TRAppDelegate.m
```diff
@@ -38,6 +40,32 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+        UILocalNotification *notification = [[UILocalNotification alloc] init];
+        notification.fireDate = [NSDate distantPast];
+        notification.alertTitle = @"Today's Weather Forecast";
+        notification.alertBody = conditions.string;
```
sharplet: We should ensure these strings are properly localizable.

Perhaps it would be worth introducing a `ForecastNotificationViewModel`?

Tropos/Controllers/TRAppDelegate.m
klaaspieter: I was hesitant to include UIKit here. 

Can you give an example of take? I'm not super familiar with RAC.

sharplet: Here's what I mean:

```diff
-    [self.applicationController.conditionsDescription subscribeNext:^(NSAttributedString *conditions) {
+    [[self.applicationController.conditionsDescription take:1] subscribeNext:^(NSAttributedString *conditions) {
```

So the signal returned by the call to `take:` will immediately complete after the first value is sent. However, on thinking about it further, there's still a chance that this signal will never complete: in the event that an error occurs, no value will be sent on the `conditionsDescription` signal. This is because it's ultimately a KVO observation of a property that only changes when the weather update is successful.

In order to ensure that the fetch as actually successful, we need to do something like the following:

```objc
    RACSignal *updatedConditions = [[self.applicationController performBackgroundFetch] then:^{
        return [self.applicationController.conditionsDescription take:1];
    }];

    [updatedConditions subscribeNext:^(NSAttributedString *conditions) {
        // schedule local notification
    }];
```

Does that make sense?

Tropos/Controllers/TRAppDelegate.m
klaaspieter: How strongly do you feel about this @sharplet. I briefly attempted to put in all of your suggestions but quickly ran into the limitations of my RAC experience. 

I feel it's ok to merge this and perhaps have someone with more RAC experience revisit later. What do you think?

PR #173, opened 2016-05-02
Create TroposCore framework and reorganise project tree

sharplet:
This is in preparation for sharing the core logic between iOS and watchOS. The models, view models and most controllers have been moved into a new `Core/` top-level directory, and the remaining app code reorganised into `Tropos/Sources/`, `Tropos/Specs/` and `Tropos/Resources/`.


gfontenot:
👍 once CI is passing

klaaspieter:
Should this be merged into #174 @sharplet ?

sharplet:
@klaaspieter If we can roll it into a beta and verify everything's working as expected I'd say :thumbsup:

PR #176, opened 2016-05-27
Fix unsubscribe from timezone

klaaspieter:
Unsubscribe from the previous channel if the device's time zone changes.

### DIFF
--- a/Tropos/Controllers/TRApplicationController.h
+++ b/Tropos/Controllers/TRApplicationController.h
@@ -13,4 +13,6 @@
 - (RACSignal *)localWeatherNotification;
 - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)application;
 
+- (void)subscribeToNotificationsWithDeviceToken:(NSData *)deviceToken;
+
 @end
--- a/Tropos/Controllers/TRApplicationController.m
+++ b/Tropos/Controllers/TRApplicationController.m
@@ -58,4 +58,20 @@
     }
 }
 
+- (void)subscribeToNotificationsWithDeviceToken:(NSData *)deviceToken
+{
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    NSString *formerChannelKey = @"CourierChannel";
+
+    NSString *channel = [[NSTimeZone localTimeZone] name];
+
+    NSString *formerChannel = [userDefaults stringForKey:formerChannelKey];
+    if (formerChannel && ![channel isEqualToString:formerChannel]) {
+        [self.courier unsubscribeFromChannel:formerChannel];
+    }
+    [self.courier subscribeToChannel:channel withToken:deviceToken];
+
+    [userDefaults setObject:channel forKey:formerChannelKey];
+}
+
 @end
--- a/UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
+++ b/UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
@@ -20,6 +20,13 @@
 
 @end
 
+void inTimeZone(NSTimeZone *timeZone, void (^callback)(void)) {
+    NSTimeZone *defaultTimeZone = [NSTimeZone defaultTimeZone];
+    [NSTimeZone setDefaultTimeZone:timeZone];
+    callback();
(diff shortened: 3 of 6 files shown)


UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
```diff
@@ -61,6 +62,44 @@ @interface TRWeatherController (Tests)
...
+            [applicationController subscribeForNotificationsWithDeviceToken:deviceToken];
+
+            OCMVerify([courier unsubscribeFromChannel:initialTimeZone.name]);
+        });
+    });
 });
 
 QuickSpecEnd
+
+void inTimeZone(NSTimeZone *timeZone, void (^callback)(void)) {
```
sharplet: Is this being used anywhere?

UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
```diff
@@ -61,6 +62,44 @@ @interface TRWeatherController (Tests)
...
+    describe(@"subscribeForNotificationsWithDeviceToken:", ^{
+        it(@"subscribes to notifications on the channel corresponding to the device's time zone", ^{
+            TRCourierClient *courier = OCMClassMock([TRCourierClient class]);
+            TRApplicationController *applicationController = [TRApplicationController new];
+            applicationController.courier = courier;
+
+            NSData *deviceToken = [NSData new];
+            [applicationController subscribeForNotificationsWithDeviceToken:deviceToken];
+
+            NSString *channel = [[NSTimeZone localTimeZone] name];
```
sharplet: What do you think about renaming this variable to `expectedChannel`, and moving its definition into the "arrange" portion of the test before the call to `subscribeForNotificationsWithDeviceToken:`?

Tropos/Controllers/TRApplicationController.h
```diff
@@ -13,4 +13,6 @@
 - (RACSignal *)localWeatherNotification;
 - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)application;
 
+- (void)subscribeForNotificationsWithDeviceToken:(NSData *)deviceToken;
```
sharplet: I think this should read `subscribeToNotifications`: generally you subscribe "to" something, not "for" it.

Tropos/Controllers/TRApplicationController.m
```diff
@@ -58,4 +58,20 @@ - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)applica
...
 
+- (void)subscribeForNotificationsWithDeviceToken:(NSData *)deviceToken
+{
+    NSUserDefaults *userDefaults = [NSUserDefaults standardUserDefaults];
+    NSString *formerChannelKey = @"CourierChannel";
+
+    NSString *channel = [[NSTimeZone localTimeZone] name];
+
+    NSString *formerChannel = [userDefaults stringForKey:formerChannelKey];
+    if (formerChannel && ![channel isEqualToString:formerChannel]) {
```
sharplet: What do you think about reversing this comparison? That way, if `formerChannel` is `nil` the expression would be a `nil` message-send, and evaluate to false, meaning that you don't need to check for `nil` first.

sharplet:
Looks good 👍 

I assume the problem here is getting notifications for multiple channels? What happens if the subscribe or unsubscribe fails? Is there any way to handle that kind of error at the moment?

klaaspieter:
@sharplet I don't think there is in this 'release' but it's something I just added. You can pass a completion handler that returns a `CourierResult` to the method. The reasoning behind not having it from the start is that it's not super useful in the subscribe case. If a subscribe failed, it'll be retried the next time an app is launched.

UnitTests/Tests/Controllers/TRApplicationControllerSpec.m
klaaspieter: It is now. I wanted to make sure any time zone changes didn't impact other tests. I'm not sure if it's actually a problem, but I hope this function prevents it either way. Plus it reads kind of nice.

Tropos/Controllers/TRApplicationController.m
klaaspieter: Thanks. It appears I forgot how to Objective-C :)

PR #177, opened 2016-05-31
Fix backgrounded notification

klaaspieter:
Don't perform background fetch when receiving push notification

When the app was launched in the background in response to a silent
notification, no local notification was ever scheduled.

Inspecting the logs revealed that an error prevented the weather
description from being updated:

```
error: scheduling: Error Domain=RACCommandErrorDomain Code=1 "The
command is disabled and cannot be executed"
UserInfo={RACUnderlyingCommandErrorKey=<RACCommand: 0x13d5d8b70>,
  NSLocalizedDescription=The command is disabled and cannot be executed}
```

It appears the `updateWeatherCommand` is being called somewhere else,
giving us this error. Because the command is never successfully
executed, no local notification is ever scheduled.


klaaspieter:
Closing PR because it should've been against `kp-release` not `master`.

PR #178, opened 2016-05-31
Don't perform background fetch when receiving push notification

klaaspieter:
When the app was launched in the background in response to a silent
notification, no local notification was ever scheduled.

Inspecting the logs revealed that an error prevented the weather
description from being updated:

```
error: scheduling: Error Domain=RACCommandErrorDomain Code=1 "The
command is disabled and cannot be executed"
UserInfo={RACUnderlyingCommandErrorKey=<RACCommand: 0x13d5d8b70>,
NSLocalizedDescription=The command is disabled and cannot be executed}
```

It appears the `updateWeatherCommand` an app coming from suspended state triggers `application:didFinishLaunching:` which in our case configures and executes the `updateWeatherCommand`.

### DIFF
--- a/Tropos/Controllers/TRApplicationController.m
+++ b/Tropos/Controllers/TRApplicationController.m
@@ -36,10 +36,7 @@
 
 - (RACSignal *)localWeatherNotification
 {
-    RACSignal *updatedConditions = [[self performBackgroundFetch] then:^{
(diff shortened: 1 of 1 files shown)


klaaspieter:
Decided to merge this because it's holding back another beta.

Issue #185, opened 2016-06-07
Not all time zones receive push notifications


klaaspieter:
Tropos broadcasts notifications to a Courier channel using the time zone name. Unfortunately the canonical names of the time zones don't always match. For example setting an iPhone to use San Francisco time zone returns “US/Pacific”. Tropos will thus [register with channel](https://github.com/thoughtbot/Tropos/blob/kp-release/Tropos/Sources/Controllers/TRApplicationController.m#L63) `us-pacific`. 

The backend uses [`ActiveSupport::Timezone`](http://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html) to determine in which time zones it's currently 9am and broadcast to those time zones. Today at 9am pacific time the backend returns: `["America/Los_Angeles", "America/Tijuana", "America/Phoenix"]` which does not include `us-pacific`.

In other words iOS is subscribing to different channels than the backend is broadcasting to. It's possible for users in these time zones to never receive notifications.

klaaspieter:
This was resolved by #187.

PR #186, opened 2016-06-08
Use GMT offset as channel name

klaaspieter:
This partially fixes the issue where certain time zones will never receive push notifications because the time zone names used by Rails are different from those used by iOS (See thoughtbot/Tropos#185 for more information)

This fixes it by not using the names, but the relative offset from UTC as the channel name. Currently we subscribe to `us-pacific` when the iOS time zone is set to San Francisco. 

This PR changes it to subscribe to the channel minus-0700 instead. This is a more reliable way of identifying time zones across platforms.


klaaspieter:
Closing because this should be against kp-release branch.

PR #187, opened 2016-06-08
Use GMT offset as Courier channel name

klaaspieter:
This partially fixes the issue where certain time zones will never receive push notifications because the time zone names used by Rails are different from those used by iOS (See thoughtbot/Tropos#185 for more information)

This fixes it by not using the names, but the relative offset from UTC as the channel name. Currently we subscribe to us-pacific when the iOS time zone is set to San Francisco.

This PR changes it to subscribe to the channel minus-0700 instead. This is a more reliable way of identifying time zones across platforms.

### DIFF
--- a/Tropos/Sources/Controllers/CourierClient.swift
+++ b/Tropos/Sources/Controllers/CourierClient.swift
@@ -1,7 +1,20 @@
 import Courier
 
+@objc protocol TimeZone {
+    var secondsFromGMT: Int { get }
+}
+extension NSTimeZone: TimeZone {}
+
 @objc(TRCourierClient) final class CourierClient: NSObject {
     private let instance: Courier
+
+    class func channelNameForTimeZone(timeZone: TimeZone) -> String {
+        let seconds = timeZone.secondsFromGMT
+        let sign = seconds < 0 ? "minus" : "plus"
+        let hours = abs(seconds) / 3600
+        let minutes = (abs(seconds) % 3600) / 60
+        return String(format: "%@ %02d%02d", sign, hours, minutes)
+    }
 
     init(apiToken: String) {
         let environment: Environment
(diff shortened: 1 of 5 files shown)


Tropos/Sources/Controllers/CourierClient.swift
```diff
@@ -1,8 +1,21 @@
...
+
 @objc(TRCourierClient) final class CourierClient: NSObject {
     private let instance: Courier
 
+    class func channelNameForTimeZone(timeZone: TimeZone) -> String {
+        let seconds = timeZone.secondsFromGMT
+        let sign = seconds < 0 ? "minus" : "plus"
+        let hours = abs(seconds) / 3600
+        let minutes = (abs(seconds) % 3600) / 60
+        return String(format: "%@ %02d%02d", sign, hours, minutes)
```
calleerlandsson: Should the space be a dash in this format string?

calleerlandsson:
I had one question, other than that this looks good to merge.

Tropos/Sources/Controllers/CourierClient.swift
klaaspieter: I wasn't sure either. Eventually this is going to be parameterized and I thought my `parameterize` implementation, which was inspired by ActiveSupport's, would do something with the dash. 

I just checked and it actually doesn't, but it also transforms spaces into dashes so the end result is the same.

calleerlandsson: Thanks for explaining! Then I guess it doesn't really matter. The reason I noticed is because I also read the server-side implementation and was expecting there to be a dash.

cpytel:
This is a good change because it also gets us closer to being able to have the user select the time for the notification.

sharplet:
Will this code handle daylight savings changes gracefully?

Tropos/Sources/Controllers/CourierClient.swift
```diff
@@ -1,8 +1,21 @@
...
+    var secondsFromGMT: Int { get }
+}
+extension NSTimeZone: TimeZone {}
+
 @objc(TRCourierClient) final class CourierClient: NSObject {
     private let instance: Courier
 
+    class func channelNameForTimeZone(timeZone: TimeZone) -> String {
+        let seconds = timeZone.secondsFromGMT
+        let sign = seconds < 0 ? "minus" : "plus"
```
jakecraige: Curious why you decided to use plus and minus strings rather than the symbols?

klaaspieter: Because both `"-7000".parameterize` and `"+7000".parameterize` return `7000`.

klaaspieter:
@cpytel hadn't thought of that, but that's absolutely true!

@sharplet there is [some logic](https://github.com/thoughtbot/Tropos/blob/kp-release/Tropos/Sources/Controllers/TRApplicationController.m#L86-L90) in place that automatically unsubscribes from a channel if the time zone changes. So assuming all goes well this deals with DST.

It'll only try once so worst case scenario it's possible for someone to be subscribe to multiple channels. I created #188 to track this issue.

PR #189, opened 2016-06-09
Correctly take temperature from yesterday's forecast

klaaspieter:


Core/Sources/Models/WeatherUpdate.swift
```diff
@@ -98,7 +98,7 @@ public extension WeatherUpdate {
     }
 
     var yesterdaysTemperature: Temperature? {
-        let rawTemperature = yesterdaysConditionsJSON["temperature"] as? Int ?? 0
+        let rawTemperature = yesterdaysConditionsJSON["currently"]?["temperature"] as? Int ?? 0
         return Temperature(fahrenheitValue: rawTemperature)
```
jakecraige: Since this is already optional, what about having it return nil if it can't get a forecast? 0 doesn't really seem like a reasonable default

PR #190, opened 2016-06-13
Fix stale notification

### LINKED ISSUE #184: Forecast in morning push notification not up-to-date
In some situations the forecast information in the morning push notification is outdated. When the app goes into suspended state it still keeps cached forecast information. The information isn't actually updated when the silent notification is received, the local notification presented to the user will show the information loaded the last time the app was in the foreground.

The refresh was disabled in #178 because it was preventing the local notification from being shown at all. The reasoning at the time was that the refresh would be triggered by `application:didFinishLaunching:` when the app was being unsuspended but that appears not to be true.

The solution seems to be trigger the `updateWeatherCommand` from `application:didReceiveRemoteNotification:completionHandler` only if it wasn't already triggered in `application:didFinishLaunching:`

klaaspieter:
By consolidating all calls to `weatherCommand execute` in the `AppDelegate` it becomes much easier to reason about when the command is executing. This made it easier to figure out at which times in the launch cycle we to execute it and how to prevent the command being executed multiple times.

The weather is currently updated when:

* in `application:didFinishLaunchingWithOptions:` if the options dictionary does not contain `UIApplicationLaunchOptionsRemoteNotificationKey`
* in `applicationWillEnterForeground`
* in `application:didReceiveRemoteNotification:fetchCompletionHandler:`

The first two replace the weather updates in `WeatherViewController` in response to `viewDidLoad` and `UIApplicationWillEnterForegroundNotification`. By moving this into the app delegate we can more easily detect when the application is launched by a remote notification and prevent any weather updates normally done to update the UI.

This PR resolves #184.


Tropos/Sources/Categories/UIApplication+TRReactiveBackgroundTask.h
```diff
@@ -0,0 +1,8 @@
+#import <UIKit/UIKit.h>
+#import <ReactiveCocoa/ReactiveCocoa.h>
+
+@interface UIApplication (TRReactiveBackgroundTask)
+
+- (RACSignal *)tr_backgroundTaskWithSignal:(RACSignal *(^)(void))signalBlock;
```
sharplet: Would you mind an attribution link to http://sharplet.me/post/123363708110/wrapping-uikit-background-tasks-with?

klaaspieter:
I'm going to merge this so I can release it in a beta.

PR #191, opened 2016-06-17
Ensure there is only one weather controller instance

klaaspieter:
This fixes issues where the weather is updated by the app delegate, but
the update isn't actually reflected in the UI.


jakecraige:
Ooh tricky 🍭

sharplet:
Dependency injection! 👍

PR #192, opened 2016-06-19
Don't archive on circleci

klaaspieter:
If the archive fails it doesn't give us any meaningful information
because actual releases are done through fastlane. It's likely we won't
pay attention to build failures if they're due to archive failures.
Subsequent failures caused by something else might be missed or ignored.


sharplet:
I believe the main reason I added this was to so that CI would have a step that tried both a Release build and a 32-bit build. Before the archive step, there was a PR where a 32-bit build failure went undetected.

PR #193, opened 2016-06-19
Indefinitely retry unsubscribing from a channel

klaaspieter:


Tropos/Sources/Controllers/CourierClient.swift
```diff
@@ -32,7 +32,7 @@ extension NSTimeZone: TimeZone {}
         instance.subscribeToChannel(channel, withToken: token)
     }
 
-    func unsubscribeFromChannel(channel: String) {
+    func unsubscribeFromChannel(channel: String, completionHandler: ((NSError?) -> Void)? = nil) {
         instance.unsubscribeFromChannel(channel)
```
sharplet: Should `completionHandler` be passed through here?

Tropos/Sources/Controllers/TRApplicationController.m
```diff
@@ -79,21 +79,31 @@ - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)applica
...
 
     NSTimeZone *timeZone = [NSTimeZone localTimeZone];
     NSString *channel = [TRCourierClient channelNameForTimeZone:timeZone];
 
-    NSString *formerChannel = [userDefaults stringForKey:formerChannelKey];
+    NSString *formerChannel = [userDefaults stringForKey:courierChannelKey];
     if (formerChannel && ![channel isEqualToString:formerChannel]) {
         NSLog(@"unsubscribe from channel: %@", formerChannel);
-        [self.courier unsubscribeFromChannel:formerChannel];
+        [[RACSignal createSignal:^RACDisposable *(id<RACSubscriber> subscriber) {
```
sharplet: Can this signal be extracted to a category/extension on `self.courier`?

Tropos/Specs/Controllers/TRApplicationControllerSpec.m
```diff
@@ -98,7 +98,11 @@ void inTimeZone(NSTimeZone *timeZone, void (^callback)(void)) {
                 [applicationController subscribeToNotificationsWithDeviceToken:deviceToken];
             });
 
-            OCMVerify([courier unsubscribeFromChannel:expectedChannel]);
+            XCTestExpectation *expectation = [self expectationWithDescription:@"Unsubscribed from channel"];
+            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 1 * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
```
sharplet: This hard-coded time delay doesn't seem like the right approach to me, as it's likely to be non-deterministic. In Kiwi you can use `stub:withBlock:` to inject a fake implementation that could be used for things like fulfilling asynchronous expectations. Does OCMock have something similar?

Tropos/Sources/Controllers/TRApplicationController.m
```diff
@@ -79,21 +79,31 @@ - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)applica
...
+            [self.courier unsubscribeFromChannel:formerChannel completionHandler:^(NSError * _Nullable error) {
+                if (!error) {
+                    [subscriber sendCompleted];
+                } else {
+                    [subscriber sendError:error];
+                }
+            }];
+
+            return [RACDisposable disposableWithBlock:^{}];
```
sharplet: You can safely return `nil` instead of a no-op disposable.

Tropos/Sources/Controllers/TRApplicationController.m
```diff
@@ -79,21 +79,31 @@ - (void)setMinimumBackgroundFetchIntervalForApplication:(UIApplication *)applica
...
+        }] retry];
```
sharplet: Because you're not actually using the signal returned from `-retry`, it looks like this code might never actually be running...

klaaspieter: Is there a `subscribeWithoutCaringAboutTheResult` type of thing?

jakecraige: In think just subscribe. At least in RxSwift, not sure about RAC but maybe?

sharplet: I generally use this:

```objc
[signal subscribeCompleted:^{
  // no-op
}];
```

sharplet: Although it's probably better to at least log the error:

```objc
[signal subscribeError:^(NSError *error) {
  NSLog(@"error: %@", error);
}];
```

PR #201, opened 2018-02-23
Bug squashing

sharplet:
Fixes an issue introduced with the Swift 3 migration, as well as a flaky test.


Tropos/Resources/Other-Sources/Info.plist
```diff
@@ -22,6 +22,19 @@
 	<string>202</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSAppTransportSecurity</key>
```
jamescmartinez: 🙈 Can't use HTTPS?

sharplet: 🤷‍♂️

PR #204, opened 2018-03-09
Remove Courier & push notifications code

sharplet:
While we had a working Courier integration, we never quite ironed out the bugs with timezone-based push notification delivery. As Courier has since been sunset, we need to remove the push notifications
functionality for now.

I've already removed the Push Notifications entitlement from the App ID.


danbee:
I'm probably not best qualified to judge but LGTM.

Tropos.xcodeproj/project.pbxproj
```diff
@@ -972,7 +957,7 @@
 								enabled = 1;
 							};
 							com.apple.Push = {
-								enabled = 1;
+								enabled = 0;
```
jamescmartinez: Nice catch. These are usually the flags I forget about.

PR #207, opened 2018-03-16
Update weather on applicationDidBecomeActive(_:)

sharplet:
I've been using Tropos with the "when in use" location authorization, and finding that sometimes I would launch the app and no weather update would occur. I _think_ the cause might be that `applicationWillEnterForeground(_:)` is called while the app is in a background state, causing the location check to fail / never complete.

Users with "always" location authorization and background fetch enabled will still get more timely weather updates.

I added some error logging, and also rewrote the app delegate in Swift, because why not!


Tropos/Sources/Controllers/TRSettingsController.m
```diff
@@ -1,75 +0,0 @@
-#import "TRSettingsController.h"
-#import "NSBundle+TRBundleInfo.h"
-
-NSString *const TRSettingsDidChangeNotification = @"TRSettingsDidChangeNotification";
-static NSString *const TRSettingsUnitSystemKey = @"TRUnitSystem";
-static NSString *const TRSettingsLastVersionKey = @"TRLastVersion";
-
-@interface TRSettingsController ()
```
sharplet: This was a duplicate of `SettingsController` as defined in the Core framework.

Tropos/Sources/main.m
```diff
@@ -1,8 +0,0 @@
-#import "TRAppDelegate.h"
```
jamescmartinez: Did you mean to remove this?

Tropos/Sources/Controllers/AppDelegate.swift
```diff
@@ -0,0 +1,70 @@
...
+
+@UIApplicationMain
+final class AppDelegate: UIResponder, UIApplicationDelegate {
+    let applicationController = ApplicationController()
+    var window: UIWindow?
+
+    func application(
+        _ application: UIApplication,
+        didFinishLaunchingWithOptions _: [UIApplicationLaunchOptionsKey: Any]? = nil
+    ) -> Bool {
```
jamescmartinez: This function is _giant_. Could it be broken down?

Tropos/Sources/main.m
sharplet: Yep, this is superseded by the `@UIApplicationMain` annotation in `AppDelegate.swift`.

Tropos/Sources/Controllers/AppDelegate.swift
sharplet: Sure, we could extract some setup functions.

sharplet: Extracted `setupHockey()` in 489f7f1

Core/Specs/Models/WeatherUpdateCacheSpec.swift
```diff
@@ -5,26 +5,31 @@ import Nimble
 private let testCacheFileName = "TestWeatherUpdate"
 
 private var testCachesDirectory: URL {
-    return FileManager.default
-        .urls(for: .documentDirectory, in: .userDomainMask)
-        .first!
+    return try! FileManager.default
```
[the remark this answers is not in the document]
sharplet: Disabled the warning for this case.

sharplet:
ugh, finally!

PR #208, opened 2018-05-18
Fix a but with precipitation probability

danbee:
It was casting to the wrong type so the text did not match the icon.

From this:
simulator screen shot - iphone x - 2018-05-18 at 13 32 07

To this:
simulator screen shot - iphone x - 2018-05-18 at 13 22 40

### DIFF
--- a/Core/Sources/Models/WeatherUpdate.swift
+++ b/Core/Sources/Models/WeatherUpdate.swift
@@ -126,7 +126,7 @@ public extension WeatherUpdate {
     }
 
     @objc var precipitationPercentage: Double {
-        return Double(todaysForecast["precipProbability"] as? String ?? "") ?? 0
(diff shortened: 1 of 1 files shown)


Core/Sources/Models/WeatherUpdate.swift
```diff
@@ -126,7 +126,7 @@ public extension WeatherUpdate {
     }
 
     @objc var precipitationPercentage: Double {
-        return Double(todaysForecast["precipProbability"] as? String ?? "") ?? 0
+        return (todaysForecast["precipProbability"] as? NSNumber)?.doubleValue ?? 0
```
sharplet: At least for consistency, I wonder if we should make a similar change (casting via `NSNumber`) for `windSpeed` and `windBearing`?

danbee: What would be the advantage of doing this?

PR #210, opened 2018-08-22
Add a basic Shortcut to have Siri read out the latest weather update

sharplet:
Screen shot of an interaction with Siri demonstrating a spoken weather update

Creates a "Check weather forecast" custom Siri intent that reads out the latest weather update from the cache. This implementation isn't quite shippable yet (weather updates will go stale very quickly), but this is the first step to having a Siri-driven workflow that asynchronously fetches a new weather update and reads it out to the user.

Check the commit log for more details!


Sources/Tropos/Extensions/INInteraction.swift
```diff
@@ -0,0 +1,16 @@
...
+extension INInteraction {
+    @objc(tr_checkWeatherInteraction)
+    static var checkWeather: INInteraction {
+        let intent = CheckWeatherIntent()
+        intent.suggestedInvocationPhrase = checkWeatherSuggestedPhrase
+        return INInteraction(intent: intent, response: nil)
+    }
+
+    private static let checkWeatherSuggestedPhrase = NSLocalizedString(
+        "Check the weather",
```
danbee: Is there any potential for this to collide with another weather app using the same phrase?

sharplet:
Regarding next steps: the intent handler will need to gain the ability to perform the async logic, which is encapsulated in the various `*Controller` classes in the iOS app target. These controllers might have to move into TroposCore so that the can be used without depending on the app UI. Ideally they'd all have been rewritten in Swift already, but it probably makes sense to port them across as-is and not wait for that rewrite to happen.

sharplet:
Ugh, finally got CI passing! Note: merging this will drop support for Xcode 9 on master, but if that becomes a problem for some reason we can address it then.

PR #213, opened 2018-08-27
Dynamically update weather forecast from Siri intent

sharplet:
Moves the `LocationController`, `GeocodeController` and `ForecastController` into TroposCore, then makes use of them in `CheckWeatherIntentHandler` to dynamically update the weather forecast before reading it to the user.


sharplet:
No idea why the tests are failing; I'm going to merge this.

PR #224, opened 2018-12-28
WIP: Remove CocoaPods

sharplet:
- [x] Switch most dependencies to Carthage
- [x] Remove OCMock altogether by refactoring ApplicationControllerSpec to use protocols and subclassing instead
- [x] Switch from Hockey (which will be discontinued this year) to App Center


bin/setup
```diff
@@ -2,12 +2,21 @@
 
 set -e
 
+gem_install_log="${GEM_INSTALL_LOG_PATH:-.gem-install.log}"
+
+gem_install() {
+  gem_name=$1
+
+  gem install "$gem_name" --user-install &> "$gem_install_log" \
+    && echo "Installed $gem_name"
```
mike-burns: Maybe `| tee $gem_install_log` instead of this dance -- then it will go to output in addition to a log. Or not, if you're doing this to keep things pretty!

sharplet: Definitely doing this to keep things pretty!

mike-burns: :nail_care:

PR #227, opened 2019-01-07
TRdailyforecast view swift refactor

katsuroo:


Sources/Tropos/Views/DailyForecastView.swift
```diff
@@ -24,12 +24,4 @@ import TroposCore
         super.layoutSubviews()
         contentView.frame = bounds
     }
-
-    func setViewModel(dailyForecastViewModel: DailyForecastViewModel) {
```
sharplet: I think we still need this method! The Objective-C version of this, `-setViewModel:`, was a method that implemented the setter for the `viewModel` property. There are a few related features in Swift, but the way to have side effects on set in Swift is with a [property observer](https://docs.swift.org/swift-book/LanguageGuide/Properties.html#ID262).

That will look something like this:

```swift
@objc public var viewModel: DailyForecastViewModel? {
  didSet {
    dayOfWeekLabel.text = viewModel?.dayOfWeek
    //  . . .
  }
}
```

Sources/Tropos/Views/DailyForecastView.swift
```diff
@@ -0,0 +1,27 @@
+import UIKit
+import TroposCore
+
+@objc(TRDailyForecastView) public class DailyForecastView: UIView {
+    @IBOutlet weak var dayOfWeekLabel: UILabel!
+    @IBOutlet weak var highTemperatureLabel: UILabel!
+    @IBOutlet weak var lowTemperatureLabel: UILabel!
+    @IBOutlet weak var conditionsImageView: UIImageView!
+    @IBOutlet var contentView: DailyForecastView!
+    @objc public var viewModel: DailyForecastViewModel!
```
sharplet: While it’s useful for the outlets to be implicitly unwrapped optionals (because they’re set automatically when the view is loaded), I think it would make sense for this `viewModel` property to be a regular optional.

Sources/Tropos/Views/DailyForecastView.swift
```diff
@@ -0,0 +1,27 @@
...
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+    }
+
+    public override func awakeFromNib() {
+        super.awakeFromNib()
+        let mainBundle = Bundle.main
+        mainBundle.loadNibNamed(String(describing: "TRDailyForecastView"), owner: self, options: nil)
```
sharplet:
```suggestion
        mainBundle.loadNibNamed(String(describing: "TRDailyForecastView"), owner: self)
```

Does this compile if you leave off the `options` parameter? I _think_ it might have a default value of `nil`.

sharplet: I think this should either be `String(describing: type(of: self))`, `String(describing: DailyForecastView.self)`, or a literal string.

Sources/Tropos/Views/DailyForecastView.swift
```diff
@@ -0,0 +1,27 @@
+import UIKit
+import TroposCore
+
+@objc(TRDailyForecastView) public class DailyForecastView: UIView {
+    @IBOutlet weak var dayOfWeekLabel: UILabel!
+    @IBOutlet weak var highTemperatureLabel: UILabel!
+    @IBOutlet weak var lowTemperatureLabel: UILabel!
+    @IBOutlet weak var conditionsImageView: UIImageView!
```
sharplet: It used to be common to mark outlets as `weak`, but these days it’s recommended to use regular strong references instead.

sharplet:
Looks like there’s some compile errors on CI:

```
[07:39:26]: ▸ ❌  /Users/distiller/project/Sources/Tropos/ViewControllers/TRWeatherViewController.m:22:38: unknown type name 'TRDailyForecastView'
[07:39:26]: ▸ @property (weak, nonatomic) IBOutlet TRDailyForecastView *oneDayForecastView;
[07:39:26]: ▸                                      ^
```

Have you been able to reproduce this locally? It’s strange, because it definitely looks like you’re importing `Tropos-Swift.h` in `TRWeatherViewController.m`.

Sources/Tropos/Views/DailyForecastView.swift
katsuroo: Did not that was not objective c setters are defined... Implemented the suggested changes to the view model setter

Sources/Tropos/Views/DailyForecastView.swift
katsuroo: Leaving out options worked.

katsuroo: I had some runtime errors when using the DailyForecastView in a description string. Wasn't sure what was happening at the time so put in a string in place of the class at the time to test but forgot to take it out. 

Just renamed the nib file to match the class name.

sharplet: > Just renamed the nib file to match the class name.

Excellent. `String(describing: DailyForecastView.self)` returns the name of the type as a string, i.e., `DailyForecastView`. `NSStringFromClass(DailyForecastView.self)` returns the Objective-C class name, which in this case is `TRDailyForecastView`, or if the name is not customised `<module>.<class>`, e.g. `Tropos.DailyForecastView`. It’s conventional to name the .xib after the class name, so `String(describing: DailyForecastView.self)` should work.

